PR #32, opened 2022-01-26
Fix payloadSize not being passed to the client

oncethor:
Without this no data is really flowing from the client to the execution
point.


Raytar:
Thanks!

Issue #34, opened 2022-02-17
Insecure implementation of BLS threshold signatures


OrestisAlpos:
The implementation of threshold signatures in `crypto/bls12/bls12/go` is insecure.
Namely, the methods `CreateThresholdSignature()` and `VerifyThresholdSignature()` actually create (and verify) an *aggregate*, not threshold, signature, which is insecure when partial signatures sign the same message. That is explained in Section 5.1 of the BLS paper: https://link.springer.com/content/pdf/10.1007/s00145-004-0314-9.pdf
What we actually want is described on Section 5.3 of the BLS paper.

The implemented version could be seen as a multi-signature, but still it is susceptible to the rogue-key attack, because it just adds the public keys and signatures. The correct implementation would be as in Section 3.1 of the BDN paper: https://eprint.iacr.org/2018/483.pdf

Even if one is only interested in benchmarking (and does not plan to use this in production), the results might still be misleading, since the computations behind a threshold signature (or a correct implementation of multi-signature) are different than what the code currently does.

Raytar:
Hi!

> Namely, the methods CreateThresholdSignature() and VerifyThresholdSignature() actually create (and verify) an aggregate, not threshold, signature, which is insecure when partial signatures sign the same message.

Indeed, the `bls12` implementation is not "production-grade" and does not protect against rogue-key attacks. The main reason I implemented it in the first place was to figure out how to support different crypto technologies in the module system, and to experiment with aggregate signatures.

> What we actually want is described on Section 5.3 of the BLS paper.

If I understand this correctly, these signatures would not validate at all unless the threshold is met, as opposed to our current implementation which verifies an aggregate and then checks that the number of participants meets the threshold?

I am currently working on implementing [Handel](https://arxiv.org/pdf/1906.05132.pdf) as an alternative way to aggregate the signatures. My implementation relies on being able to verify an aggregate signature even before it meets the threshold for consensus. Would this be possible with true threshold signatures?

Raytar:
> I wonder if we should at least rename `CreateThresholdSignature` and `VerifyThresholdSignature` to `CreateMultiSignature`and`VerifyMultiSignature`.

I discussed this issue with @meling yesterday, and he suggested refactoring the `CryptoImpl` interface such that we have only three methods: `Sign`, `Verify`, and `Combine`. The `Verify` and `Combine` methods should be smart enough to accept both single signatures and multi/aggregate/threshold signatures depending on the implementation. That way we can avoid these naming problems.

[...]

PR #36, opened 2022-02-28
Crypto refactor

johningve:
This is a complete rewrite of the crypto implementations based on discussion in #34.

# Important Changes:

* The `CryptoImpl` interface is replaced with the `CryptoBase` interface which is much simpler:
```go
// CryptoBase provides the basic cryptographic methods needed to create, verify, and combine signatures.
type CryptoBase interface {
	// Sign creates a cryptographic signature of the given message.
	Sign(message []byte) (signature QuorumSignature, err error)
	// Combine combines multiple signatures into a single signature.
	Combine(signatures ...QuorumSignature) (signature QuorumSignature, err error)
	// Verify verifies the given quorum signature against the message.
	Verify(signature QuorumSignature, message []byte) bool
	// BatchVerify verifies the given quorum signature against the batch of messages.
	BatchVerify(signature QuorumSignature, batch map[hotstuff.ID][]byte) bool
}
```
* The `ThresholdSignature` interface is renamed to `QuorumSignature`, and the `Signature` interface is removed.
* `ConnectionMetadata` allows for metadata to be sent to other replicas.
* This metadata is used by the `bls12` package to send a proof-of-possession to all replicas.


Raytar:
@meling There are some changes to various tests because I had to change the `testutil.TestModules` function and make the `testutil.CreateBuilders` use the twins network to get the new metadata working. It's kinda ugly now, but I'll probably revisit it at some point and try to clean it up.

backend/config.go
```diff
@@ -110,6 +116,50 @@ func NewConfig(creds credentials.TransportCredentials, opts ...gorums.ManagerOpt
...
+	for k, v := range m {
+		md.Set(keyPrefix+k, v)
+	}
+	return md
+}
+
+func readMetadata(md metadata.MD) map[string]string {
+	m := make(map[string]string)
+	for k, values := range md {
+		if strings.HasPrefix(k, keyPrefix) && len(values) > 0 {
```
meling, line 156: Go 1.18 has a new `strings.Cut` function that could replace both `HasPrefix` and `TrimPrefix`. Something like this, I think:
```go
if _, key, ok := strings.Cut(k, keyPrefix); ok && len(values) > 0 {
  m[key] = values[0]
```
Maybe the most important place to do this, but if done elsewhere as well, maybe `Cut` can be more efficient.

backend/config.go
```diff
@@ -75,17 +84,14 @@ type Config struct {
...
-	// embed own ID to allow other replicas to identify messages from this replica
-	md := metadata.New(map[string]string{
-		"id": fmt.Sprintf("%d", cfg.mods.ID()),
+	cfg.mods.EventLoop().RegisterHandler(replicaConnected{}, func(event interface{}) {
+		if !cfg.connected {
+			cfg.mods.EventLoop().DelayUntil(connected{}, event)
+			return
+		}
+		c := event.(replicaConnected)
+		cfg.replicaConnected(c)
```
meling, line 93:
```go
cfg.replicaConnected(event.(replicaConnected))
```

backend/server.go
```diff
@@ -35,9 +34,11 @@ func (srv *Server) InitConsensusModule(mods *consensus.Modules, _ *consensus.Opt
 func NewServer(opts ...gorums.ServerOption) *Server {
 	srv := &Server{}
 
-	grpcServerOpts := []grpc.ServerOption{}
-
-	opts = append(opts, gorums.WithGRPCServerOptions(grpcServerOpts...))
+	// grpcServerOpts := []grpc.ServerOption{}
+	// opts = append(opts, gorums.WithGRPCServerOptions(grpcServerOpts...))
```
meling, line 38: Are these two lines no longer necessary?

consensus/cryptoiface.go
```diff
@@ -0,0 +1,82 @@
...
+func VerifyMulti(messages map[hotstuff.ID][]byte) VerifyOption {
+	return func(o *VerifyOptions) {
+		o.MultipleMessages = true
+		o.MessageMap = messages
+	}
+}
+
+// VerifyOptions specify options for the Verify function in the CryptoBase interface.
+type VerifyOptions struct {
+	// using a pointer for the Message such that we can tell if it is unspecified
```
meling, line 54: I think `Message` does not need to be a pointer. You can still distinguish these two cases, if needed:
```go
b = []byte{} // non-nil, but len(b) == 0
b = nil // nil and len(b) == 0
```
I tried to remove the pointer and update the relevant code and the tests in the `crypto` package still pass.

consensus/cryptoiface.go
```diff
@@ -0,0 +1,82 @@
...
+
+	Message          *[]byte
+	MessageMap       map[hotstuff.ID][]byte
+	MultipleMessages bool // true if `MessageMap` should be used instead of `Message`
+	Threshold        int
+}
+
+// Crypto implements the methods required to create and verify signatures and certificates.
+// This is a higher level interface that is implemented by the crypto package itself.
+type Crypto interface {
```
meling, line 64: I'm not a fan of big interfaces like this. Any reason to have all these methods in the same interface?

If you make separate interfaces for each method or pairs of methods, it would be easier to implement alternatives. Although, I'm not sure we will have multiple implementations of these methods.

consensus/types.go
```diff
@@ -110,33 +103,43 @@ type PrivateKey interface {
...
 
+// ThresholdSignature is a signature that is only valid when it contains the signatures of a quorum of replicas.
+//
+// Deprecated: renamed to QuorumSignature
+type ThresholdSignature = QuorumSignature
+
 // PartialCert is a signed block hash.
 type PartialCert struct {
-	signature Signature
+	// shortcut to the signer of the signature
```
meling, line 120: In general, I'm not sure the `consensus` package is the right place for these certificate types. I wonder if it wouldn't be better to place them in `crypto` or `crypto/certificates` or something along these lines.

consensus/cryptoiface.go
```diff
@@ -0,0 +1,82 @@
+package consensus
+
+import (
+	"github.com/relab/hotstuff"
+)
+
+// CryptoBase provides the basic cryptographic methods needed to create, verify, and combine signatures.
+type CryptoBase interface {
```
meling, line 8: See the naming/composition here: [`Signer`](https://pkg.go.dev/crypto#Signer). There is no interface that includes a verify function. On the other hand, it seems the std lib doesn't define an interface that's used across signature types. Instead most signing/verify algorithms just use functions specific for each signature type.

meling: I guess my point was that I don't think `CryptoBase` is the right name for this interface, since the methods are related to digital signatures in some form. The other question is whether or not we should use an interface.

meling: Similar to other question: why is this in `consensus`?

Raytar: I haven't yet found a better place to put these types/interfaces without import cycles. It would be best to keep them in the crypto package, but I also need the interfaces to be available in the module system. And the crypto package needs the module system...

consensus/cryptoiface.go
Raytar: The reason we have this interface is to avoid an import cycle. The only implementor is the `crypto` struct in `crypto/crypto.go`, and I don't see any reason for alternative implementations. But the methods of the `crypto` struct are needed by the other modules. So we can have this interface, or we would have to move the whole `crypto` struct into the `consensus` package as well. Maybe there is a better way, but I haven't been able to think of one yet.

backend/config.go
meling: Suggest to use `Cut`, since we probably want to switch go Go 1.18 soon anyway.

backend/config.go
```diff
@@ -110,6 +115,50 @@ func NewConfig(creds credentials.TransportCredentials, opts ...gorums.ManagerOpt
...
+	}
+
+	replica.(*Replica).md = readMetadata(md)
+
+	cfg.mods.Logger().Debugf("Replica %d connected from address %v", id, info.Addr)
+}
+
+const keyPrefix = "hotstuff-"
+
+func convertMetadata(m map[string]string) metadata.MD {
```
meling, line 144: Maybe rename this to `mapToMetadata`

backend/config.go
```diff
@@ -119,6 +168,18 @@ type ReplicaInfo struct {
 
 // Connect opens connections to the replicas in the configuration.
 func (cfg *Config) Connect(replicas []ReplicaInfo) (err error) {
+	opts := *cfg.optsPtr
+	cfg.optsPtr = nil // we don't need to keep the options around beyond this point, so we'll allow them to be GCed.
```
meling, line 172: Is this an important point? I mean is this causing memory leaks or something? Maybe it would be better to make a copy of options and let that be GCed when exiting the method. I mean `Connect` should be called rarely, right?

backend/config.go
```diff
@@ -150,6 +212,9 @@ func (cfg *Config) Connect(replicas []ReplicaInfo) (err error) {
 		replica.node = node
 	}
 
+	cfg.connected = true
+	cfg.mods.EventLoop().AddEvent(connected{})
```
meling, line 216: Would be nice with a comment referring back to the `connected{}` event handler, and what it will do on this event.

consensus/cryptoiface.go
meling: We should ideally refactor things to avoid such strong coupling, but that should be done as a separate PR.

consensus/cryptoiface.go
```diff
@@ -0,0 +1,82 @@
...
+func VerifyThreshold(threshold int) VerifyOption {
+	return func(o *VerifyOptions) {
+		o.Threshold = threshold
+	}
+}
+
+// VerifySingle verifies the signature against the given message.
+//
+// NOTE: VerifySingle and VerifyMulti cannot be used together,
+// and whichever option is specified last will be the effective one.
```
meling, line 33: If they cannot be used together, would it not be better to panic, since it would be a programmer mistake to combine them??

consensus/cryptoiface.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+}
+
+// VerifyOptions specify options for the Verify function in the CryptoBase interface.
+type VerifyOptions struct {
+	// using a pointer for the Message such that we can tell if it is unspecified
+
+	Message          *[]byte
+	MessageMap       map[hotstuff.ID][]byte
+	MultipleMessages bool // true if `MessageMap` should be used instead of `Message`
```
meling, line 58: Couldn't we encode the single message variant in the `MessageMap` with `hotstuff.ID = 0`?? Then we could avoid the `Message` field and probably the `MultipleMessages` field as well.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+type bls12Base struct {
 	mods *consensus.Modules
-}
 
-// New returns a new bls12-381 signer and verifier.
-func New() consensus.CryptoImpl {
-	bc := &bls12Crypto{}
-	return bc
+	mut      sync.RWMutex
+	popCache map[string]bool
```
meling, line 138: What is `popCache` storing? Add comment?

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
-// Sign signs a hash.
-func (bc *bls12Crypto) Sign(hash consensus.Hash) (sig consensus.Signature, err error) {
-	p, err := bls12.NewG2().HashToCurve(hash[:], domain)
-	if err != nil {
-		return nil, fmt.Errorf("bls12: hash to curve failed: %w", err)
-	}
-	pk := bc.getPrivateKey()
-	bls12.NewG2().MulScalarBig(p, p, pk.p)
-	return &Signature{signer: bc.mods.ID(), s: p}, nil
+func (bls *bls12Base) getPrivateKey() *PrivateKey {
```
meling, line 158: `s/getPrivateKey/privateKey/g`??

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+}
+
+func (bls *bls12Base) checkPop(replica consensus.Replica) (valid bool) {
+	defer func() {
+		if !valid {
+			bls.mods.Logger().Debugf("Invalid proof-of-possession for replica %d", replica.ID())
+		}
+	}()
+
+	b, ok := replica.Metadata()[popMetadataKey]
```
meling, line 217: What is `b`? A more descriptive variable name would help.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+	_, _ = key.Write(replica.PublicKey().(*PublicKey).ToBytes())
+
+	bls.mut.RLock()
+	valid, ok = bls.popCache[key.String()]
+	bls.mut.RUnlock()
+	if ok {
+		return valid
+	}
+
+	p, err := bls12.NewG2().FromCompressed([]byte(b))
```
meling, line 234: What is `p`? More descriptive variable name.

crypto/bls12/bls12.go
```diff
@@ -86,35 +91,6 @@ func (priv *PrivateKey) Public() consensus.PublicKey {
...
-	}
-	return nil
-}
-
-// Signer returns the ID of the replica that generated the signature.
-func (s *Signature) Signer() hotstuff.ID {
-	return s.signer
-}
-
 // AggregateSignature is a bls12-381 aggregate signature. The participants map contains the IDs of the replicas who
```
meling, line 94: ... The participants bitmap contains... the replicas that participated in signature creation.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
 }
 
-// TODO: I'm not sure to what extent we are vulnerable to a rogue public key attack here.
-// As far as I can tell, this is not a problem right now because we do not yet support reconfiguration,
-// and all public keys are known by all replicas.
+func (bls *bls12Base) popProve() *bls12.PointG2 {
+	pubKey := bls.getPrivateKey().Public().(*PublicKey)
+	proof, err := bls.coreSign(pubKey.ToBytes(), domainPOP)
+	if err != nil {
+		bls.mods.Logger().Panicf("failed to generate proof of possession: %v", err)
```
meling, line 201: Should logged messages be capitalized? `"Failed to ..."`

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+	bls.mut.Lock()
+	bls.popCache[key.String()] = valid
+	bls.mut.Unlock()
+
+	return valid
+}
+
+func (bls *bls12Base) coreAggregateVerify(publicKeys []*PublicKey, messages [][]byte, signature *bls12.PointG2) bool {
+	n := len(publicKeys)
+	if n != len(messages) && n <= 1 {
```
meling, line 250: Do you need `&& n <= 1`? Does this method not work for `n=1` and `len(publicKeys) == len(messages)`? Maybe it would help to add a comment to explain the limitations of this method.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+	var aggregate bls12.PointG1
+	for _, pk := range publicKeys {
+		engine.G1.Add(&aggregate, &aggregate, pk.p)
+	}
+	return bls.coreVerify(&PublicKey{p: &aggregate}, message, signature, domain)
+}
+
+// Sign creates a cryptographic signature of the given messsage.
+func (bls *bls12Base) Sign(message []byte) (signature consensus.QuorumSignature, err error) {
+	// TODO: consider passing message to this function instead of hash
```
meling, line 291: Still valid TODO? The input `message` seems to indicate it is already a message.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
-		replica, ok := bc.mods.Configuration().Replica(id)
+
+	var (
+		publicKeys = make([]*PublicKey, 0, l)
+		messages   = make([][]byte, 0, l)
+		valid      = true
+	)
+
+	// get all public keys and messages to verify
+	sig.participants.RangeWhile(func(i hotstuff.ID) bool {
```
meling, line 332: `s/i/id/g`

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+		replica, ok := bls.mods.Configuration().Replica(i)
 		if !ok {
+			valid = false
 			return false
 		}
-		pk, ok := replica.PublicKey().(*PublicKey)
-		if !ok {
+
+		if replica.ID() != bls.mods.ID() && !bls.checkPop(replica) {
```
meling, line 339: A comment on this line would help.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,257 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+			valid = false
 			return false
 		}
-		p2, err := bls12.NewG2().HashToCurve(hash[:], domain)
-		if err != nil {
-			return false
+		publicKeys = append(publicKeys, replica.PublicKey().(*PublicKey))
+
+		if opts.MultipleMessages {
```
meling, line 345: If `MultipleMessages` is false, then `messages` above is not used.

Overall, it seems like the code could be simplified by avoiding the separation between single message verify functions and multi-message verify functions.

crypto/cache.go
```diff
@@ -3,36 +3,27 @@ package crypto
...
-	impl        consensus.CryptoImpl
+	impl        consensus.CryptoBase
 	mut         sync.Mutex
 	capacity    int
-	entries     map[key]*list.Element
+	entries     map[string]*list.Element
 	accessOrder list.List
 }
 
 // NewCache returns a new Crypto implementation that caches the results of the operations of the given CryptoImpl
```
meling, line 20: `s/CryptoImpl/CryptoBase/`

crypto/cache.go
```diff
@@ -193,7 +130,8 @@ func (cache *cache) VerifyThresholdSignatureForMessageSet(signature consensus.Th
 // As opposed to the CreateThresholdSignature methods,
 // this method does not check whether the resulting
 // signature meets the quorum size.
-func (cache *cache) Combine(signatures ...interface{}) consensus.ThresholdSignature {
+func (cache *cache) Combine(signatures ...consensus.QuorumSignature) (consensus.QuorumSignature, error) {
 	// we don't cache the result of this operation, because it is not guaranteed to be valid.
-	return cache.impl.Combine(signatures...)
+	sig, err := cache.impl.Combine(signatures...)
```
meling, line 135: Since `err` isn't checked, you can simply return directly.

crypto/cache.go
```diff
@@ -73,117 +63,64 @@ func (cache *cache) evict() {
...
 	}
-	key := cache.accessOrder.Remove(cache.accessOrder.Back()).(key)
+	key := cache.accessOrder.Remove(cache.accessOrder.Back()).(string)
 	delete(cache.entries, key)
 }
 
-// Sign signs a hash.
-func (cache *cache) Sign(hash consensus.Hash) (sig consensus.Signature, err error) {
-	sig, err = cache.impl.Sign(hash)
+// Sign signs a message.
```
meling, line 70: ... and adds it to the cache.

crypto/cache.go
```diff
@@ -73,117 +63,64 @@ func (cache *cache) evict() {
...
+func (cache *cache) Sign(message []byte) (sig consensus.QuorumSignature, err error) {
+	sig, err = cache.impl.Sign(message)
 	if err != nil {
 		return nil, err
 	}
-	key := sha256.Sum256(append(hash[:], sig.ToBytes()...))
-	cache.insert(key, false)
+	var key strings.Builder
+	hash := sha256.Sum256(message)
+	_, _ = key.Write(hash[:])
```
meling, line 78: Suggest to add a comment to explain why we ignore checking these errors...

crypto/cache.go
```diff
@@ -73,117 +63,64 @@ func (cache *cache) evict() {
...
-	}
-	var key consensus.Hash
-	hash := sha256.New()
-	for _, h := range hashes {
-		hash.Write(h[:])
+	var hash consensus.Hash
+	// if we have multiple messages, we will hash them all together.
+	if opts.MultipleMessages {
+		hasher := sha256.New()
+		for _, m := range opts.MessageMap {
```
meling, line 103: Won't this result in non-deterministic iteration ordering?

crypto/cache.go
```diff
@@ -193,7 +130,8 @@ func (cache *cache) VerifyThresholdSignatureForMessageSet(signature consensus.Th
 // As opposed to the CreateThresholdSignature methods,
```
meling, line 130: Check this documentation.. including ^^ Should it be `single threshold`?

crypto/crypto.go
```diff
@@ -6,112 +6,120 @@ import (
...
-	sig, err := base.CreateThresholdSignature(sigs, view.ToHash())
+	sig, err := c.Combine(sigs...)
 	if err != nil {
 		return consensus.TimeoutCert{}, err
 	}
 	return consensus.NewTimeoutCert(sig, view), nil
 }
 
-func (base base) CreateAggregateQC(view consensus.View, timeouts []consensus.TimeoutMsg) (aggQC consensus.AggregateQC, err error) {
+func (c crypto) CreateAggregateQC(view consensus.View, timeouts []consensus.TimeoutMsg) (aggQC consensus.AggregateQC, err error) {
```
meling, line 72: Missing doc

meling: Is there a linter that could check this. Probably `golangci-lint` has one, but didn't check.

crypto/crypto.go
```diff
@@ -6,112 +6,120 @@ import (
...
 	}
-	sigs := make([]consensus.Signature, 0, len(timeouts))
+	sigs := make([]consensus.QuorumSignature, 0, len(timeouts))
 	for _, timeout := range timeouts {
 		sigs = append(sigs, timeout.ViewSignature)
 	}
-	sig, err := base.CreateThresholdSignature(sigs, view.ToHash())
+	sig, err := c.Combine(sigs...)
 	if err != nil {
 		return consensus.TimeoutCert{}, err
```
meling, line 67: We should avoid returning empty structs when we have error; it causes unnecessary allocation.

crypto/crypto.go
```diff
@@ -6,112 +6,120 @@ import (
...
 	}
-	sigs := make([]consensus.Signature, 0, len(signatures))
+	sigs := make([]consensus.QuorumSignature, 0, len(signatures))
 	for _, sig := range signatures {
 		sigs = append(sigs, sig.Signature())
 	}
-	sig, err := base.CreateThresholdSignature(sigs, block.Hash())
+	sig, err := c.Combine(sigs...)
 	if err != nil {
 		return consensus.QuorumCert{}, err
```
meling, line 50: We should avoid returning empty structs when we have error; it causes unnecessary allocation.

crypto/crypto.go
```diff
@@ -6,112 +6,120 @@ import (
...
 	}
 }
 
 // CreatePartialCert signs a single block and returns the partial certificate.
-func (base base) CreatePartialCert(block *consensus.Block) (cert consensus.PartialCert, err error) {
-	sig, err := base.Sign(block.Hash())
+func (c crypto) CreatePartialCert(block *consensus.Block) (cert consensus.PartialCert, err error) {
+	sig, err := c.Sign(block.ToBytes())
 	if err != nil {
 		return consensus.PartialCert{}, err
```
meling, line 33: We should avoid returning empty structs when we have error; it causes unnecessary allocation.

crypto/crypto.go
```diff
@@ -6,112 +6,120 @@ import (
...
 		}
 		if timeout.MsgSignature != nil {
 			sigs = append(sigs, timeout.MsgSignature)
-			hashes[timeout.ID] = timeout.Hash()
 		}
 	}
-	sig, err := base.CreateThresholdSignatureForMessageSet(sigs, hashes)
+	sig, err := c.Combine(sigs...)
 	if err != nil {
 		return aggQC, err
```
meling, line 85: We should avoid returning non-nil structs when we have error...

crypto/crypto.go
```diff
@@ -6,112 +6,120 @@ import (
...
+	if !ok {
+		return false
+	}
+	return c.Verify(cert.Signature(), consensus.VerifySingle(block.ToBytes()))
 }
 
 // VerifyQuorumCert verifies a quorum certificate.
-func (base base) VerifyQuorumCert(qc consensus.QuorumCert) bool {
+func (c crypto) VerifyQuorumCert(qc consensus.QuorumCert) bool {
 	if qc.BlockHash() == consensus.GetGenesis().Hash() {
```
meling, line 101: Perhaps add comment here:
```go
// genesis QC is always valid.
```

crypto/crypto.go
```diff
@@ -6,112 +6,120 @@ import (
...
+	if !ok {
+		return false
+	}
+	return c.Verify(qc.Signature(), consensus.VerifySingle(block.ToBytes()))
 }
 
 // VerifyTimeoutCert verifies a timeout certificate.
-func (base base) VerifyTimeoutCert(tc consensus.TimeoutCert) bool {
+func (c crypto) VerifyTimeoutCert(tc consensus.TimeoutCert) bool {
 	if tc.View() == 0 {
```
meling, line 113: Perhaps add comment here too.

crypto/crypto.go
```diff
@@ -121,17 +129,17 @@ func (base base) VerifyAggregateQC(aggQC consensus.AggregateQC) (bool, consensus
...
 			ID:       id,
 			View:     aggQC.View(),
 			SyncInfo: consensus.NewSyncInfo().WithQC(qc),
-		}.Hash()
+		}.ToBytes()
 	}
-	ok := base.VerifyThresholdSignatureForMessageSet(aggQC.Sig(), hashes)
+	ok := c.Verify(aggQC.Sig(), consensus.VerifyMulti(messages))
 	if !ok {
 		return false, consensus.QuorumCert{}
```
meling, line 140: Should probably return `nil, false`...

crypto/bls12/bls12.go
meling: I assume this will panic. What's the policy for logging a panic vs emitting it to stdout; I assume this is the difference... There are other places where regular panic is used.

crypto/bls12/bls12.go
Raytar: Oops, this is actually wrong. n >= 1 is a precondition of the function, and the length of the public keys and messages slices must be equal

meling (changes requested):
Looks nice! Mainly, I would prefer to have the verification functions work on a slice or map instead of having to decide via a boolean which path to take; this would also avoid two fields when you only need one of them... if necessary, the bool could be replaced with a `len(x) == 1` check.

crypto/crypto.go
Raytar: I thought that the memory would be allocated on the stack since it is returned by value?

crypto/crypto.go
Raytar: We have a linter for that, but it doesn't care about this method, because it is defined on an unexported struct.

crypto/cache.go
Raytar: It is because writing to a string builder never returns an error. It is just immplementing the io.Writer interface.

crypto/crypto.go
meling: I have gotten into the habit of always returning a pointer type in such scenarios. I'm not sure if it always doesn't escape to the heap, but it probably doesn't matter much in terms of performance. I had to look it up, and it seems both are okay, but it seems the [recommended pattern](https://morioh.com/p/385f6171e967) is to use the `nil` variant. This seems to be based on the advice on [Declaring Empty Slices](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices), which also references a talk about nil that I've watched before but didn't rewatch.

I guess one argument for using a `nil` pointer is that it becomes easier to identify/see that it is in fact returning a `nil` as opposed to examining the code and seeing `consensus.QuorumCert{}` you sort of have to think what it means that this is non-nil.

consensus/cryptoiface.go
meling: Marking this as resolved since I created an issue for the possible refactor of the module system.

consensus/types.go
meling: Marking this as resolved since I created an issue for the possible refactor of the module system.

crypto/cache.go
meling: Reasonable. Probably fine to not add a comment then.

crypto/bls12/bls12.go
Raytar: Actually, I don't have a policy for that.

Here are some more differences when compared with `panic`:

`logging.Panic` and `logging.Panicf`:
- can only be used where the logger module is accessible
- the message will be logged even if the panic is recovered
- `Panicf` supports string formatting

crypto/bitfield.go
```diff
@@ -9,18 +9,21 @@ type Bitfield struct {
...
+	if !bf.isSet(byteIdx, bitIdx) {
+		bf.len++
+	}
+	bf.data[byteIdx] |= 1 << bitIdx
 }
 
-func (bm Bitfield) isSet(byteIdx, bitIdx int) bool {
-	return bm.data[byteIdx]&(1<<bitIdx) != 0
+func (bf Bitfield) isSet(byteIdx, bitIdx int) bool {
+	return bf.data[byteIdx]&(1<<bitIdx) != 0
```
meling, line 26: The id and index methods below should use shift `>> 3` and `<< 3`. Although not sure if the compiler replaces `*8` and `/8` accordingly...

crypto/bls12/bls12.go
```diff
@@ -318,12 +318,16 @@ func (bls *bls12Base) Verify(signature consensus.QuorumSignature, options ...con
 		opt(&opts)
 	}
 
-	if !opts.MultipleMessages && opts.Message == nil {
-		panic("no hash(es) to verify the signature against: you must specify one of the VerifyHash or VerifyHashes options")
+	if len(opts.Messages) == 0 {
+		panic("no message(s) to verify the signature against: you must specify one of the VerifyHash or VerifyHashes options")
```
meling, line 322: `s/VerifyHash/VerifySingle`
`s/VerifyHashes/VerifyMulti`

crypto/bitfield.go
Raytar: Why?

meling: Well, I guess it is only necessary/useful if the compiler doesn't already replace Divide and Multiply operations with shift operations, which are substantially faster. Doing it ourselves makes it fast for sure, independent of compiler optimizations, but I guess it makes the code slightly less readable. (I guess a comment can alleviate the latter.)

meling: On second thought, maybe it is better to leave it as is for clarity... For fun, I studied the assembly code generated by these operations. It is actually shorter code with the `/` and `*` operations, but only because of the `%` operation. That is, the simple `* 8` does become a shift `<< 3`. But the `/ 8` doesn't become only a shift operation, when compiled standalone. However, since the function in question has both `/ 8` and `% 8`, it seems to be able to reuse the parts that aren't shift... (but the whole thing is non-trivial to really understand...) The main takeaway is that compilers are smart and we probably should do less manual optimization unless it is demonstrably faster in benchmarks.

crypto/crypto.go
Raytar: I just don't think that it is a good idea for constructor functions to return a pointer unless it is needed for some specific purpose. What good could a pointer do in this case?

crypto/bls12/bls12.go
Raytar: The problem with removing this separation, i.e. using the map for both `VerifySingle` and `VerifyMulti` is that it becomes difficult to check which option was used. For example, if the length is 1 and there is only one participant in the signature, then you still don't know if `VerifySingle` or `VerifyMulti` was used.

meling: Okay.

meling:
We should make some progress towards getting this merged. What remains to be discussed and what remains to be implemented?

johningve:
@meling I split the Verify function into two `Verify` and `BatchVerify`. I think this is more sensible than the approach of using the `VerifyOption` stuff. These functions are no longer responsible for checking whether the number of signatures constitutes a quorum, I moved that to the higher-level `Crypto` module's functions.

backend/config.go
```diff
@@ -52,16 +55,22 @@ func (r *Replica) NewView(msg consensus.SyncInfo) {
...
+	return r.md
+}
+
 // Config holds information about the current configuration of replicas that participate in the protocol,
 // and some information about the local replica. It also provides methods to send messages to the other replicas.
 type Config struct {
-	mods    *consensus.Modules
-	optsPtr *[]gorums.ManagerOption // using a pointer so that options can be GCed after initialization
+	mods      *consensus.Modules
+	opts      []gorums.ManagerOption // using a pointer so that options can be GCed after initialization
```
meling, line 72: This no longer uses a pointer to the slice; maybe the comment should be removed or rephrased?

backend/config.go
```diff
@@ -75,17 +84,16 @@ type Config struct {
...
 	cfg.mods = mods
 
-	opts := *cfg.optsPtr
-	cfg.optsPtr = nil // we don't need to keep the options around beyond this point, so we'll allow them to be GCed.
-
-	// embed own ID to allow other replicas to identify messages from this replica
-	md := metadata.New(map[string]string{
-		"id": fmt.Sprintf("%d", cfg.mods.ID()),
+	// This receives `replicaConnected` events from the server.
+	// We can only process these events after the configuration has been connected,
```
meling, line 88: The comment was a bit longwinded. Suggest rephrasing, e.g.,:
```
We delay processing `replicaConnected` events until after the configuration's `connected` event has occurred.
```

backend/config.go
```diff
@@ -119,15 +171,28 @@ type ReplicaInfo struct {
 
 // Connect opens connections to the replicas in the configuration.
 func (cfg *Config) Connect(replicas []ReplicaInfo) (err error) {
+	opts := cfg.opts
+	cfg.opts = nil // we don't need to keep the options around beyond this point, so we'll allow them to be GCed.
```
meling, line 175:
```go
// options are not needed beyond this point, so we allow them to be GCed.
```

crypto/bls12/bls12.go
```diff
@@ -86,37 +91,8 @@ func (priv *PrivateKey) Public() consensus.PublicKey {
...
-}
-
-// Signer returns the ID of the replica that generated the signature.
-func (s *Signature) Signer() hotstuff.ID {
-	return s.signer
-}
-
-// AggregateSignature is a bls12-381 aggregate signature. The participants map contains the IDs of the replicas who
-// participated in the creation of the signature. This allows us to build an aggregated public key to verify the
+// AggregateSignature is a bls12-381 aggregate signature. The participants bitmap contains the IDs of the replicas that
```
meling, line 94: Just to clarify, do you mean `The participant's bitmap` or the plural form??

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,294 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+type bls12Base struct {
 	mods *consensus.Modules
-}
 
-// New returns a new bls12-381 signer and verifier.
-func New() consensus.CryptoImpl {
-	bc := &bls12Crypto{}
-	return bc
+	mut sync.RWMutex
+	// popCache caches the results of popVerify for each public key.
```
meling, line 147: Should mention here and probably other places that `pop` = `proof of possession`

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,294 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
 }
 
-// TODO: I'm not sure to what extent we are vulnerable to a rogue public key attack here.
-// As far as I can tell, this is not a problem right now because we do not yet support reconfiguration,
-// and all public keys are known by all replicas.
+func (bls *bls12Base) popProve() *bls12.PointG2 {
+	pubKey := bls.privateKey().Public().(*PublicKey)
+	proof, err := bls.coreSign(pubKey.ToBytes(), domainPOP)
+	if err != nil {
+		bls.mods.Logger().Panicf("Failed to generate proof of possession: %v", err)
```
meling, line 225: Should be `proof-of-possession` or Lines 237 and 243 should be changed.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,294 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+}
+
+func (bls *bls12Base) popVerify(pubKey *PublicKey, proof *bls12.PointG2) bool {
+	return bls.coreVerify(pubKey, pubKey.ToBytes(), proof, domainPOP)
+}
+
+func (bls *bls12Base) checkPop(replica consensus.Replica) (valid bool) {
+	defer func() {
+		if !valid {
+			bls.mods.Logger().Debugf("Invalid proof-of-possession for replica %d", replica.ID())
```
meling, line 237: is this debug or warn?

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,294 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+				}
+				participants.Add(id)
+				return true
+			})
+			if err != nil {
+				return nil, err
+			}
+			g2.Add(&agg, &agg, &sig2.sig)
+		} else {
+			bls.mods.Logger().Panicf("cannot combine signature of incompatible type %T (expected %T)", sig1, sig2)
```
meling, line 354: `Cannot combine incompatible signature type ...` (same applies to Line 364.)

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,294 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+}
+
+// Verify verifies the given quorum signature against the message.
+func (bls *bls12Base) Verify(signature consensus.QuorumSignature, message []byte) bool {
+	s, ok := signature.(*AggregateSignature)
+	if !ok {
+		bls.mods.Logger().Panicf("cannot verify signature of incompatible type %T (expected %T)", signature, s)
+	}
+
+	l := s.Participants().Len()
```
meling, line 367: Previously `n` was used for a similar length. Suggest using `n` instead, because `l` looks almost like `1`.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,294 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
-		sigs[sig.Signer()] = s
+		return bls.coreVerify(pk, message, &s.sig, domain)
 	}
-	if len(sigs) < bc.mods.Configuration().QuorumSize() {
-		return nil, multierr.Combine(crypto.ErrNotAQuorum, err)
+
+	// else:
+	pks := make([]*PublicKey, 0, l)
+	s.Participants().ForEach(func(id hotstuff.ID) {
+		pk, ok := bls.publicKey(id)
```
meling, line 382: Can be moved into the `if` since `pk` isn't used later in the block.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,294 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
+			bls.mods.Logger().Warnf("Missing public key for ID %d", id)
+			return false
 		}
-		sigs[sig.Signer()] = s
+		return bls.coreVerify(pk, message, &s.sig, domain)
 	}
-	if len(sigs) < bc.mods.Configuration().QuorumSize() {
-		return nil, multierr.Combine(crypto.ErrNotAQuorum, err)
+
+	// else:
```
meling, line 379: Add more context; assume you referring to the `if l == 1`, meaning that `l > 1`.

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,294 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
-// different message hash.
-func (bc *bls12Crypto) CreateThresholdSignatureForMessageSet(partialSignatures []consensus.Signature, hashes map[hotstuff.ID]consensus.Hash) (consensus.ThresholdSignature, error) {
-	// Don't care about the hashes for signature aggregation.
-	return bc.CreateThresholdSignature(partialSignatures, consensus.Hash{})
-}
+// BatchVerify verifies the given quorum signature against the batch of messages.
+func (bls *bls12Base) BatchVerify(signature consensus.QuorumSignature, batch map[hotstuff.ID][]byte) bool {
+	s, ok := signature.(*AggregateSignature)
+	if !ok {
+		bls.mods.Logger().Panicf("cannot verify signature of incompatible type %T (expected %T)", signature, s)
```
meling, line 399: `Cannot verify incompatible signature type`

crypto/bls12/bls12.go
```diff
@@ -155,231 +131,294 @@ func (agg AggregateSignature) Bitfield() crypto.Bitfield {
...
-		return nil, multierr.Combine(crypto.ErrNotAQuorum, err)
+
+	// else:
+	pks := make([]*PublicKey, 0, l)
+	s.Participants().ForEach(func(id hotstuff.ID) {
+		pk, ok := bls.publicKey(id)
+		if ok {
+			pks = append(pks, pk)
+		} else {
+			bls.mods.Logger().Warnf("Missing public key for ID %d", id)
```
meling, line 386: Does it make sense to continue the loop in this case? Ref. to a similar construction on Line 413.

crypto/cache.go
```diff
@@ -73,127 +65,134 @@ func (cache *cache) evict() {
...
 }
 
-// Combine combines multiple signatures into a single threshold signature.
-// Arguments can be singular signatures or threshold signatures.
-//
-// As opposed to the CreateThresholdSignature methods,
-// this method does not check whether the resulting
-// signature meets the quorum size.
-func (cache *cache) Combine(signatures ...interface{}) consensus.ThresholdSignature {
+// Verify verifies a signature given a hash.
```
meling, line 143: Do we still need this commented code?

crypto/cache.go
```diff
@@ -3,36 +3,29 @@ package crypto
...
+	impl        consensus.CryptoBase
 	mut         sync.Mutex
 	capacity    int
-	entries     map[key]*list.Element
+	entries     map[string]*list.Element
 	accessOrder list.List
 }
 
-// NewCache returns a new Crypto implementation that caches the results of the operations of the given CryptoImpl
+// NewCache returns a new Crypto implementation that caches the results of the operations of the given CryptoBase
```
meling, line 22: NewCache returns a new Crypto instance that caches the results of the operations of the provided CryptoBase.

crypto/cache.go
```diff
@@ -73,127 +65,134 @@ func (cache *cache) evict() {
...
-	hash := sha256.New()
-	for _, h := range hashes {
-		hash.Write(h[:])
-	}
-	hash.Write(signature.ToBytes())
-	hash.Sum(key[:0])
-	if cache.check(key, true) {
+// BatchVerify verifies the given quorum signature against the batch of messages.
+func (cache *cache) BatchVerify(signature consensus.QuorumSignature, batch map[hotstuff.ID][]byte) bool {
+	// get a sorted list of the ids in the Messages map
```
meling, line 107:
```
// sort the list of ids from the batch map
```

crypto/cache.go
```diff
@@ -73,127 +65,134 @@ func (cache *cache) evict() {
...
+	ids := make([]hotstuff.ID, 0, len(batch))
+	for id := range batch {
```
meling, line 109: Can't recall having seen this construction before to sort the map keys, but it looks more efficient than to first loop over the map to create a slice before sorting. That said, I wonder if it would be better to make this a function called `SortedIDs` that takes a map with `hotstuff.ID` and possibly any type (when we use generics -- assuming we need to sort IDs from maps with other types.) 

For the record, there isn't coming a `SortedKeys` in the stdlib yet, but maybe in [the future.](https://github.com/golang/go/issues/53196)

crypto/cache.go
```diff
@@ -73,127 +65,134 @@ func (cache *cache) evict() {
...
+		ids = append(ids, 0)
+		copy(ids[i+1:], ids[i:])
+		ids[i] = id
+	}
+
+	var hash consensus.Hash
+	hasher := sha256.New()
+	// then hash the messages in sorted order
+	for _, id := range ids {
+		m := batch[id]
```
meling, line 122: Can replace `m` with `batch[id]` directly in the `Write` call.

crypto/crypto.go
```diff
@@ -0,0 +1,153 @@
...
+	sig, err := c.Combine(sigs...)
+	if err != nil {
+		return consensus.AggregateQC{}, err
+	}
+	return consensus.NewAggregateQC(qcs, sig, view), nil
+}
+
+// VerifyPartialCert verifies a single partial certificate.
+func (c crypto) VerifyPartialCert(cert consensus.PartialCert) bool {
+	block, ok := c.mods.BlockChain().Get(cert.BlockHash())
```
meling, line 93: Wouldn't it be better to pass in `block` to the function, instead of accessing the blockchain in the crypto package, causing an implicit dependency?  Same applies to Line 109. Maybe this is fixed in #61?

crypto/crypto.go
```diff
@@ -0,0 +1,153 @@
...
+	return c.Verify(cert.Signature(), block.ToBytes())
+}
+
+// VerifyQuorumCert verifies a quorum certificate.
+func (c crypto) VerifyQuorumCert(qc consensus.QuorumCert) bool {
+	// genesis QC is always valid.
+	if qc.BlockHash() == consensus.GetGenesis().Hash() {
+		return true
+	}
+	if qc.Signature().Participants().Len() < c.mods.Configuration().QuorumSize() {
```
meling, line 106: Similar to my [comment above](https://github.com/relab/hotstuff/pull/36/files#r888374558), the quorum size or Configuration could be passed in, which would decouple the crypto package from the module system.

At least my initial reaction was that this may cause problems due to tight coupling.

crypto/crypto.go
```diff
@@ -0,0 +1,153 @@
...
+		messages[id] = consensus.TimeoutMsg{
+			ID:       id,
+			View:     aggQC.View(),
+			SyncInfo: consensus.NewSyncInfo().WithQC(qc),
+		}.ToBytes()
+	}
+	if aggQC.Sig().Participants().Len() < c.mods.Configuration().QuorumSize() {
+		return consensus.QuorumCert{}, false
+	}
+	ok = c.BatchVerify(aggQC.Sig(), messages)
```
meling, line 145: Move into if?

meling: Actually, this code (Lines 145-152) could be written as:
```go
if c.BatchVerify(aggQC.Sig(), messages) && c.VerifyQuorumCert(highQC) {
	return highQC, true
}
return consensus.QuorumCert{}, false
```
Perhaps adding a comment above the if to say that both must verify...

crypto/ecdsa/ecdsa.go
```diff
@@ -127,225 +123,146 @@ func (sig ThresholdSignature) RangeWhile(f func(hotstuff.ID) bool) {
...
+	s, ok := signature.(MultiSignature)
+	if !ok {
+		ec.mods.Logger().Panicf("cannot verify signature of incompatible type %T (expected %T)", signature, s)
+	}
 
-		hash, ok := hashes[s.Signer()]
-		if !ok {
-			continue
-		}
+	l := signature.Participants().Len()
```
meling, line 198: `s/l/n/g`

meling: The above comment meant: substitute `l` with `n`.

crypto/ecdsa/ecdsa.go
```diff
@@ -127,225 +123,146 @@ func (sig ThresholdSignature) RangeWhile(f func(hotstuff.ID) bool) {
...
 	}
-	return numVerified >= ec.mods.Configuration().QuorumSize()
-}
 
-// VerifyThresholdSignatureForMessageSet verifies a threshold signature against a set of message hashes.
-func (ec *ecdsaCrypto) VerifyThresholdSignatureForMessageSet(signature consensus.ThresholdSignature, hashes map[hotstuff.ID]consensus.Hash) bool {
-	ec.mods.Logger().Debug(hashes)
-	sig, ok := signature.(ThresholdSignature)
-	if !ok {
+	l := signature.Participants().Len()
```
meling, line 229: `s/l/n/g`

crypto/ecdsa/ecdsa.go
```diff
@@ -127,225 +123,146 @@ func (sig ThresholdSignature) RangeWhile(f func(hotstuff.ID) bool) {
...
+
+	results := make(chan bool, l)
+
+	set := make(map[string]struct{})
+	for id, sig := range s {
+		message, ok := batch[id]
 		if !ok {
 			return false
 		}
+		set[string(message)] = struct{}{}
```
meling, line 242: This might be costly for long messages. Perhaps it would be better to stringify the `hash` and use that as the map key?? It would be interesting to benchmark this. @hanish520

meling (changes requested):
Looks great! There are a few minor things, and a few things that need some discussion. I don't want to hold it back longer, so the things to be discussed can also be postponed for another PR, if we decide they would help decoupling.

I noticed tests are failing due to golangci-lint, as you already noticed in #58; probably replacing [this line](https://github.com/relab/hotstuff/blob/crypto-refactor/.github/workflows/golangci-lint.yml#L17) with 1.18 would help.

crypto/bls12/bls12.go
johningve: I mean the `participants` field in the struct.

crypto/bls12/bls12.go
johningve: Yeah, will do. Here I would prefer to have an explicit else clause, because I think it makes the code clearer, but it goes against Go's style (and linters will complain).

crypto/bls12/bls12.go
johningve: `ok` is checked by the if statement though.

crypto/bls12/bls12.go
johningve: No it does not. Will change.

crypto/cache.go
johningve: I played around with some different ways to make this generic. However, I think that that using the generic `maps.Keys` function from the experimental `maps` package to extract the keys to a slice, and then the generic `slices.Sort` function from the experimental `slices` package to sort the keys is easier and likely faster.

johningve: Here are some benchmark results:

```
cpu: AMD Ryzen 9 5900HS with Radeon Graphics        
BenchmarkKeysAsc-16      	       4	 282953742 ns/op	 2238214 B/op	     979 allocs/op
BenchmarkSliceSort-16    	     104	  11535986 ns/op	  858424 B/op	      42 allocs/op
```

The first one, `KeysAsc`, uses the sort.Search method in the same way as the code above.
The second one, `SliceSort`, first uses `maps.Keys` and then sorts them with `slices.Sort`.

Both are sorting the keys of a map with 100 000 entries

johningve: When sorting maps with only 10 entries, the `slices.Sort` version still wins:

```
cpu: AMD Ryzen 9 5900HS with Radeon Graphics        
BenchmarkKeysAsc-16      	 5160612	       230.5 ns/op	      80 B/op	       1 allocs/op
BenchmarkSliceSort-16    	 7871254	       152.8 ns/op	      80 B/op	       1 allocs/op
```

crypto/crypto.go
johningve: I guess it is a responsibility problem. Should the caller of `VerifyPartialCert` be responsible for finding the block, or should the crypto implementation itself do it? I prefer it the way it is because it makes the crypto interface easier to use.

crypto/crypto.go
johningve: I think that passing in the quorum size or configuration complicates the interface. I don't think the Crypto module's dependency on the BlockChain module is a problem.

crypto/ecdsa/ecdsa.go
johningve: Actually, we can just use the hash itself. No need to stringify

crypto/ecdsa/ecdsa.go
johningve: Indeed, I am familiar with sed syntax ;)

PR #37, opened 2022-02-28
Handel

johningve:
Full implementation of [Handel](https://arxiv.org/abs/1906.05132)

Depends on #36


Raytar:
@Arian7394 I think it is working now, thanks to your suggestions :)

Raytar:
It does seem like multiple levels can be active at the same time, based on these sentences from the paper:

>A node sends a message to one node from each of its active levels. Levelℓ is active if Outℓ is complete or if StartTimeℓ has elapsed.

But maybe it isn't necessary as you mentioned.

PR #57, opened 2022-05-23
eventloop: make it an interface

johningve:
This adds an interface for the event loop, allowing for alternative implementations or overriding methods.

I added the interface in order to override the `AddEvent` method for the Twins nodes, but I no longer have a need for that. However, I think it might be useful to have this interface in case it becomes useful later.

### DIFF
--- a/eventloop/eventloop.go
+++ b/eventloop/eventloop.go
@@ -13,14 +13,53 @@ import (
 	"time"
 )
 
-// EventHandler processes an event.
-type EventHandler func(event interface{})
-
 // EventLoop accepts events of any type and executes relevant event handlers.
 // It supports registering both observers and handlers based on the type of event that they accept.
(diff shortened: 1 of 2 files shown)


meling:
Have done a quick high-level scan of this PR. My general comment is that the number of methods in the interface is larger than typical Go interfaces. Also, it is recommended practice to return structs from constructor functions instead of an interface. Will try to get some time to look more carefully at the whole code base before making recommendations.

meling:
I'm closing this PR since it goes against the idiomatic Go style of keeping interfaces small, and interfaces are usually not returned from (constructor) functions but instead accepted as input to functions. Interfaces should be defined where they are used, not upfront, just in case it becomes useful.

The book 100 Go Mistakes and How to Avoid Them provides some good guidelines for use of interfaces in items
- 5: Interface pollution,
- 6: Interface on the producer side, and
- 7: Returning interfaces

PR #60, opened 2022-05-27
modules/registry: use generics to constrain types

johningve:
We can use generics to constrain the types of module constructors and pointers at compile time to avoid having to check them at runtime.

### DIFF
--- a/internal/orchestration/worker.go
+++ b/internal/orchestration/worker.go
@@ -163,25 +163,24 @@ func (w *Worker) createReplica(opts *orchestrationpb.ReplicaOpts) (*replica.Repl
 	// prepare modules
 	builder := consensus.NewBuilder(hotstuff.ID(opts.GetID()), privKey)
 
-	var consensusRules consensus.Rules
-	if !modules.GetModule(opts.GetConsensus(), &consensusRules) {
+	consensusRules, ok := modules.GetModule[consensus.Rules](opts.GetConsensus())
+	if !ok {
 		return nil, fmt.Errorf("invalid consensus name: '%s'", opts.GetConsensus())
 	}
 
-	var byz byzantine.Byzantine
-	if modules.GetModule(opts.GetByzantineStrategy(), &byz) {
+	if byz, ok := modules.GetModule[byzantine.Byzantine](opts.GetByzantineStrategy()); ok {
 		consensusRules = byz.Wrap(consensusRules)
 	} else if opts.GetByzantineStrategy() != "" {
 		return nil, fmt.Errorf("invalid byzantine strategy: '%s'", opts.GetByzantineStrategy())
 	}
 
-	var cryptoImpl consensus.CryptoImpl
-	if !modules.GetModule(opts.GetCrypto(), &cryptoImpl) {
+	cryptoImpl, ok := modules.GetModule[consensus.CryptoImpl](opts.GetCrypto())
+	if !ok {
 		return nil, fmt.Errorf("invalid crypto name: '%s'", opts.GetCrypto())
 	}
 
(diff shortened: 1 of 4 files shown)


internal/orchestration/worker.go
```diff
@@ -163,25 +163,24 @@ func (w *Worker) createReplica(opts *orchestrationpb.ReplicaOpts) (*replica.Repl
...
+	consensusRules, ok := modules.GetModule[consensus.Rules](opts.GetConsensus())
+	if !ok {
 		return nil, fmt.Errorf("invalid consensus name: '%s'", opts.GetConsensus())
 	}
 
-	var byz byzantine.Byzantine
-	if modules.GetModule(opts.GetByzantineStrategy(), &byz) {
+	if byz, ok := modules.GetModule[byzantine.Byzantine](opts.GetByzantineStrategy()); ok {
 		consensusRules = byz.Wrap(consensusRules)
 	} else if opts.GetByzantineStrategy() != "" {
```
meling, line 173: Don't you want to err out `if !ok`? If I understand this, it means an unknown/invalid byzantine strategy was supplied.

modules/registry.go
```diff
@@ -12,18 +12,12 @@ var (
...
 //  RegisterModule("chainedhotstuff", func() consensus.Rules { return chainedhotstuff.New() })
-func RegisterModule(name string, constructor interface{}) {
-	ctorType := reflect.TypeOf(constructor)
-
-	if ctorType.Kind() != reflect.Func && ctorType.NumOut() != 1 && ctorType.Out(0).Kind() != reflect.Interface {
-		panic("invalid argument: constructor must be a function returning an interface")
-	}
-
-	ifaceType := ctorType.Out(0)
+func RegisterModule[T any](name string, constructor func() T) {
```
meling, line 18: I have a question. Is it useful to allow two or more modules (names) that have the same `T`? I couldn't see any uses that requires more than one module per `T`. In which case, there isn't really a need for the `name` argument, except to allow string-based lookup (via the untyped version). I think the latter could be handled if we assume that the `name` is the same as the package...

meling (commented):
Looks good. I have a few questions that could be discussed...

modules/registry.go
meling: Would something like this work?
```go
	byName      = make(map[string]any)
)

func RegisterModule[T any](constructor func() T) {
	var t T
	moduleType := reflect.TypeOf(t)
	name := moduleType.PkgPath()
	name = name[strings.LastIndex(name, "/")+1:]

	registryMut.Lock()
	defer registryMut.Unlock()

	byName[name] = constructor
}

func GetModule[T any]() (out T, ok bool) {
	targetType := reflect.TypeOf(out)
	name := targetType.PkgPath()
	name = name[strings.LastIndex(name, "/")+1:]

	registryMut.Lock()
	defer registryMut.Unlock()

	ctor, ok := byName[name]
	if !ok {
		return out, false
	}

	return ctor.(func() T)(), true
}
```

johningve: Yes, we do want to be able to register multiple modules per type `T`. For example, for the interface type `consensus.Rules` we want to register our three different implementations, `chainedhotstuff`, `fasthotstuff`, and `simplehotstuff`. The registry allows us to construct these implementations at runtime by providing the name of the implementation.

internal/orchestration/worker.go
johningve: This is indeed a bit confusing. I have refactored it. What really should happen is this:

1. If `ByzantineStrategy` is empty, do nothing
1. If `GetModule` succeeds, wrap the consensus rules
1. Else return an error

PR #63, opened 2022-06-05
Initial commit for cpu stats

hanish520:
Arian wanted to see the amount of CPU and memory saved if some of the replicas chose not to participate in the Handel. I thought maybe it is interesting in general hotstuff experiments to understand the CPU and memory usage. 
__*Did not clean up the code, and please let me know if the direction is fine.*__


metrics/replicacpumem.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// InitModule gives the module access to the other modules.
+func (c *CPUMemStat) InitModule(mods *modules.Modules) {
+	c.mods = mods
+	c.mods.EventLoop().RegisterObserver(types.TickEvent{}, func(event interface{}) {
+		c.tick(event.(types.TickEvent))
+	})
+	c.mods.Logger().Info("CPU-Memory stats metric enabled")
+	// Percent with 0 interval returns 0 usage when called first time.
```
johningve, line 39: I think this comment should be more descriptive. Maybe:
```
The cpu.Percent function returns the CPU usage since the last call when called with an interval of 0.
This initial call ensures that the first measurement of the CPU usage is nonzero.
```

metrics/replicacpumem.go
```diff
@@ -0,0 +1,83 @@
...
+	_, err := cpu.Percent(0, false)
+	if err != nil {
+		c.mods.Logger().Info("Unable to fetch the CPU usage")
+	}
+}
+
+// getCPUsage Method returns the average CPU per core and the number of cores, including logical ones.
+func (c *CPUMemStat) getCPUsage() (float64, uint32) {
```
johningve, line 47: I would rename this method to `getCPUPercentage`.

metrics/replicacpumem.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return 0, 0
+	}
+	return v.Available, v.UsedPercent
+}
+
+// tick method is invoked periodically based on the configured measuring interval of metrics
+func (c *CPUMemStat) tick(_ types.TickEvent) {
+	now := time.Now()
+	cpusage, cores := c.getCPUsage()
```
johningve, line 73: Make sure to use `camelCase` for variable names. Change to `cpuUsage`, `availableMem`, `memUsage`

metrics/types/types.proto
```diff
@@ -39,3 +39,11 @@ message ViewTimeouts {
   // Number of view timeouts.
   uint64 Timeouts = 3;
 }
+
+message CPUMemoryStats {
+  Event Event = 1;
+  double CPUsagePercentage = 2;
```
johningve, line 45: Either change CPUsagePercentage to CPU**U**sagePercentage or drop "Usage" from both.
I.E. `CPUPercentage` and `MemoryPercentage`

johningve (changes requested):
This is quite neat. I have some nitpicks about variable names but otherwise, it looks good.

metrics/types/types.proto
```diff
@@ -39,3 +39,11 @@ message ViewTimeouts {
   // Number of view timeouts.
   uint64 Timeouts = 3;
 }
+
+message CPUMemoryStats {
```
meling, line 43: Suggest renaming this to `CPUMemoryMeasurement` to align with the `Throughput` and `Latency` variants.

metrics/replicacpumem.go
```diff
@@ -0,0 +1,84 @@
...
+func init() {
+	RegisterReplicaMetric("cpumem", func() interface{} {
+		return &CPUMem{}
+	})
+	RegisterClientMetric("cpumem", func() interface{} {
+		return &CPUMem{}
+	})
+}
+
+// CPUMem measures CPU usage and Memory usage and record in the metric logs.
```
meling, line 27: CPUMem measures CPU and memory usage and records these in the metrics logs.

metrics/replicacpumem.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"time"
+
+	"github.com/relab/hotstuff/metrics/types"
+	"github.com/relab/hotstuff/modules"
+	"github.com/shirou/gopsutil/v3/cpu"
+	"github.com/shirou/gopsutil/v3/mem"
+)
+
+// CPUMem metics measures the percentage of cpu and memory utilization on the node.
```
meling, line 12: This is kind of important info for a user. Please move this doc to the `metrics` package level, and integrate it with other documentation for the other metrics supported.

Note: Some typos in this text: `metic(s)` -> `metric(s)`

metrics/replicacpumem.go
```diff
@@ -0,0 +1,84 @@
+package metrics
```
meling, line 1: Since this file covers both `replica` and `client` CPU and memory metrics, I suggest renaming the file to `cpumem.go` instead.

metrics/replicacpumem.go
```diff
@@ -0,0 +1,84 @@
...
+	c.mods.Logger().Info("CPU-Memory stats metric enabled")
+	// The cpu.Percent function returns the CPU usage since the last call when called with an interval of 0.
+	// This initial call ensures that the first measurement of the CPU usage is nonzero.
+	_, err := cpu.Percent(0, false)
+	if err != nil {
+		c.mods.Logger().Info("Unable to fetch the CPU usage")
+	}
+}
+
+// getCPUPercentage Method returns the average CPU per core and the number of cores, including logical ones.
```
meling, line 47:
```go
// getCPUPercentage returns the average CPU per core and the number of logical cores.
```

metrics/replicacpumem.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return 0, 0
+	}
+	return v.Available, v.UsedPercent
+}
+
+// tick method is invoked periodically based on the configured measuring interval of metrics
+func (c *CPUMem) tick(_ types.TickEvent) {
+	now := time.Now()
+	cpu, cores := c.getCPUPercentage()
```
meling, line 74: I think we don't need the two helper methods `getCPUPercentage()` and `getMemoryPercentage()`. The code would be easier to read if they were just inlined here.

metrics/replicacpumem.go
```diff
@@ -0,0 +1,84 @@
...
+	v, err := mem.VirtualMemory()
+	if err != nil {
+		return 0, 0
+	}
+	return v.Available, v.UsedPercent
+}
+
+// tick method is invoked periodically based on the configured measuring interval of metrics
+func (c *CPUMem) tick(_ types.TickEvent) {
+	now := time.Now()
```
meling, line 73: `now` is only used on line 77. You can just call `time.Now()` there.

metrics/replicacpumem.go
```diff
@@ -0,0 +1,84 @@
...
+	mods *modules.Modules
+}
+
+// InitModule gives the module access to the other modules.
+func (c *CPUMem) InitModule(mods *modules.Modules) {
+	c.mods = mods
+	c.mods.EventLoop().RegisterObserver(types.TickEvent{}, func(event interface{}) {
+		c.tick(event.(types.TickEvent))
+	})
+	c.mods.Logger().Info("CPU-Memory stats metric enabled")
```
meling, line 38:
```go
CPU-Memory metric enabled
```

meling (changes requested):
This has been here for a long time, and I finally reviewed it... 👍🏼 

Anyway, I think it is better to split this into two separate metrics, one for memory and another for CPU. While they might be used together, they require two different "calls" to the runtime API, and thus it doesn't seem logical to me to combine them into one. Also, it prevents you from enabling only one of them. It also makes naming things more awkward.

meling:
Could we drop the dependency in favor of the new metrics API in stdlib. See [runtime/metrics](https://pkg.go.dev/runtime/metrics). Or maybe this makes the whole PR obsolete?

PR #72, opened 2022-08-02
Fix go version

AeonSw4n:
Hey there! Started playing around with your repo today and noticed it requires go 1.18 while the docs mention go 1.16. Here's a quick fix.


johningve:
Thanks!

PR #73, opened 2022-08-12
Refactor module system

johningve:


johningve:
Note to self: Squash this PR when merging

johningve:
@meling I decided to drop the explicit `modules.Implements[T]` declaration for now. The rule now is that `mods.Get` will give the last compatible value that was added to the modules builder.

johningve:
@meling I have myself reviewed all the changes and am happy with what is here. If you think that the changes to the module system look okay, then I think that we can merge it

modules/core.go
```diff
@@ -23,133 +23,145 @@
...
-}
+// TryGet attempts to find a module for ptr.
+// TryGet returns true if a module was stored in ptr, false otherwise.
+//
+// NOTE: ptr must be a non-nil pointer to a type that has been provided to the module system.
+//
+// Example:
+//
+//	builder := modules.New()
+//	builder.Provide(MyModuleImpl{}, new(MyModule))
```
meling, line 64: Should this be `Add`?

modules/core.go
```diff
@@ -23,133 +23,145 @@
...
-// Deprecated: The metrics event loop is no longer separate from the main event loop. Use EventLoop() instead.
-func (mods Core) MetricsEventLoop() *eventloop.EventLoop {
-	return mods.EventLoop()
+// NOTE: ptr must be a non-nil pointer to a type that has been provided to the module system.
+// Get panics if ptr is not a pointer, or if a compatible module is not found.
+//
+// Example:
+//
+//	builder := modules.New()
+//	builder.Provide(MyModuleImpl{}, new(MyModule))
```
meling, line 100: Should this be `Add`?

modules/core.go
```diff
@@ -23,133 +23,145 @@
...
-	}
-	targetType := outType.Elem()
-	if m, ok := mods.modulesByType[targetType]; ok {
-		reflect.ValueOf(dest).Elem().Set(reflect.ValueOf(m))
-		return true
+// GetAll finds a module for all the given pointers.
+//
+// NOTE: pointers must only contain non-nil pointers to types that have been provided to the module system.
+// GetAll panics if one of the given pointers is not a pointer, or if a compatible module is not found.
+func (mods *Core) GetAll(pointers ...any) {
```
meling, line 115: Suggest the remove `GetAll` and migrate the `...any` to the `Get` method, so that you can call `Get` with one or many arguments. Zero arguments is a no-op so there is no need to panic, but it is a programmer mistake, so could also panic.

docs/modules.md
```diff
@@ -0,0 +1,286 @@
...
+But later, when we want to implement on a more dynamic leader rotation scheme,
+we realize that we need historical information about the previous views and thus need to interact with the synchronizer and blockchain modules.
+The synchronizer probably needs to know about the identity of the leader, so we now have to refactor these somehow.
+
+It seems then, that strict adherence to the best practice of avoiding circular dependencies is only going to complicate the matter of creating new module implementations.
+Of course, we should keep this best practice in mind, but we should allow circular dependencies in order to make it easier to implement new modules.
+
+### Event-based Indirection
+
+This is not a good solution for reasons we'll get to later, but it warrants discussion anyway.
```
meling, line 63: What is "this" referring to?

docs/modules.md
```diff
@@ -0,0 +1,286 @@
...
+The idea is to use the event system that we have developed to facilitate communication between modules.
+Instead of each module having a direct dependency on other modules,
+it would depend upon the event system to deliver some request to another module.
+The other module would then process the request and emit an event containing the result.
+
+The problems of this approach are fairly obvious.
```
meling, line 69: What is "this" ref to?

PR #74, opened 2022-08-19
Implement event-based contexts for view changes and timeouts

johningve:
This PR implements two functions, `ViewContext` and `TimeoutContext`. They return contexts that can be used for operations that should be canceled upon reaching a specific view, or when a timeout occurs.

Implementing these contexts require changes to the event loop. Specifically, we have to reintroduce asynchronous handlers, as the contexts must be able to react to new events to cancel tasks that may be running on the event loop itself. I chose to refactor the event loop handlers by adding some handler options. Currently, the only options are `RunAsync` and `WithPriority`. `RunAsync` makes the handler execute as part of `AddEvent`. `WithPriority` effectively turns the handler into an "observer". The idea is that handlers that don't require much time to run can use the `WithPriority` option to be executed before the other handlers. I am deprecating the Observer functionality in favor of the priority options.


eventloop/eventloop.go
```diff
@@ -82,19 +198,21 @@ loop:
 			el.startTicker(ctx, e.tickerID)
```
meling, line 198: I think `ctx` can be removed from the `startTicker` method since it can be accessed via the `el.ctx` field (via the `Context()` method).

eventloop/eventloop.go
```diff
@@ -103,41 +221,75 @@ func (el *EventLoop) Tick() bool {
 	if e, ok := event.(startTickerEvent); ok {
 		el.startTicker(context.Background(), e.tickerID)
```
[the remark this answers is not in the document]
meling: Related to my comment on Line 198.

eventloop/eventloop.go
meling: The `ctx` could possibly also be removed from the `runTicker()` method, and move the fetching of the `Context()` into `runTicker()`. But not sure it is so important.

However, I observe that `startTicker` already holds the same lock as `Context()`, and so it could simply do:
```go
var ctx context.Context
ctx, ticker.cancel = context.WithCancel(el.ctx)
```

meling (changes requested):
Mostly nice, but please check my comments. We may need to discuss some of them.

meling (approved):

PR #75, opened 2022-08-19
Optimize memory

johningve:
Various memory optimizations.

Before:
[profile (1).pb.gz](https://github.com/relab/hotstuff/files/9382654/profile.1.pb.gz)

After:
[profile.pb.gz](https://github.com/relab/hotstuff/files/9382651/profile.pb.gz)


meling:
@johningve Could you resolve the conflicts in this PR before I review it?

meling:
I looked at this, and I like many of the changes here independent of the memory improvement gained.

It probably does not make sense to try to salvage this PR as is, but perhaps we should start a new PR and redo this on top of master. @johningve Can you recall why this never moved out of draft mode?

PR #76, opened 2022-09-23
Removed leafBlock from synchronizer

leandernikolaus:
I wanted to study the semantics of the synchronizer, and I found that the `leafBlock` is always the block from the `highQC`.
Therefore, in the current implementation, we could omit the `leafBlock`.

### DIFF
--- a/synchronizer/synchronizer.go
+++ b/synchronizer/synchronizer.go
@@ -26,7 +26,6 @@ type Synchronizer struct {
 	currentView hotstuff.View
 	highTC      hotstuff.TimeoutCert
 	highQC      hotstuff.QuorumCert
-	leafBlock   *hotstuff.Block
 
 	// A pointer to the last timeout message that we sent.
 	// If a timeout happens again before we advance to the next view,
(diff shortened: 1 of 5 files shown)


synchronizer/synchronizer.go
```diff
@@ -355,14 +348,8 @@ func (s *Synchronizer) updateHighQC(qc hotstuff.QuorumCert) {
 		return
 	}
 
-	oldBlock, ok := s.blockChain.Get(s.highQC.BlockHash())
-	if !ok {
-		s.logger.Panic("Block from the old highQC missing from chain")
-	}
-
-	if newBlock.View() > oldBlock.View() {
+	if newBlock.View() > s.highQC.View() {
```
leandernikolaus, line 351: I believe this change should be done, even when not removing `leafBlock`.

meling:
I've looked at the code, and it looks fine to me. But could edit the PR description to provide some more context for this. Do you think we won't need the `LeafBlock()` for alternative synchronizer implementations?

meling (approved):
LGTM!

meling:
I don't see any issues with this PR.

@hanish520 Would this conflict with any of your other changes?

@johningve If you have a cycle to spare, please let us know your take.

I want to merge this if there are no concerns with this.

PR #78, opened 2022-10-03
Pipelining

leandernikolaus:
This implements pipelining for chained hotstuff.
Pipelined blocks still form a single chain, but need not include the certificate for their direct parent.

Commit rule is updated to commit requests if there is a depth 3 confirmation chain and all intermediate blocks are confirmed as well.

### DIFF
--- a/synchronizer/synchronizer.go
+++ b/synchronizer/synchronizer.go
@@ -36,7 +36,7 @@ type Synchronizer struct {
 	duration ViewDuration
 	timer    *time.Timer
 
-	// This should be retrieved from a configuration
+	// number of cuncurrent views, 0 and 1 both give no concurrency.
 	pipelinedViews uint8
 
 	viewCtx   context.Context // a context that is cancelled at the end of the current view
@@ -94,7 +94,7 @@ func New(viewDuration ViewDuration) modules.Synchronizer {
 		leafBlock:   hotstuff.GetGenesis(),
 		currentView: 1,
 
-		pipelinedViews: 2,
+		pipelinedViews: 2, // this should be retrieved from a configuration.
 
 		viewCtx:   ctx,
 		cancelCtx: cancel,
@@ -123,6 +123,14 @@ func (s *Synchronizer) Start(ctx context.Context) {
 	if s.currentView == 1 && s.leaderRotation.GetLeader(s.currentView) == s.opts.ID() {
 		s.consensus.Propose(s.SyncInfo())
 	}
+}
+
+// PipelinedViews returns 1 or the number of concurrent views in the pipeline
+func (s *Synchronizer) PipelinedViews() hotstuff.View {
+	if s.pipelinedViews < 1 {
+		return 1
+	}
+	return hotstuff.View(s.pipelinedViews)
 }
 
 // HighQC returns the highest known QC.
--- a/synchronizer/synchronizer.go
+++ b/synchronizer/synchronizer.go
@@ -36,6 +36,9 @@ type Synchronizer struct {
 	duration ViewDuration
 	timer    *time.Timer
 
+	// This should be retrieved from a configuration
+	pipelinedViews uint8
+
 	viewCtx   context.Context // a context that is cancelled at the end of the current view
 	cancelCtx context.CancelFunc
 
@@ -90,6 +93,8 @@ func New(viewDuration ViewDuration) modules.Synchronizer {
 	return &Synchronizer{
 		leafBlock:   hotstuff.GetGenesis(),
 		currentView: 1,
+
+		pipelinedViews: 0,
 
 		viewCtx:   ctx,
 		cancelCtx: cancel,
@@ -192,6 +197,11 @@ func (s *Synchronizer) OnLocalTimeout() {
 	s.consensus.StopVoting(s.currentView)
 
 	s.configuration.Timeout(timeoutMsg)
+
+	if s.currentView < s.highQC.View()+hotstuff.View(s.pipelinedViews) {
+		s.AdvanceView(hotstuff.NewSyncInfo().WithTimeoutView(s.currentView))
+	}
+
 	s.OnRemoteTimeout(timeoutMsg)
 }
 
@@ -315,6 +325,26 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
 		}
 	}
 
+	b, ok := syncInfo.Block()
+	if ok {
+		highQCBlock, ok := s.blockChain.Get(s.highQC.BlockHash())
+		if !ok {
+			s.logger.Error("AdvanceView: Could not find block referenced by new QC!")
+		}
+		if !s.blockChain.Extends(b, highQCBlock) {
+			b = highQCBlock
+		}
+		s.leafBlock = b
+		if b.View() > v && b.View() < v+hotstuff.View(s.pipelinedViews) {
+			v = b.View()
+		}
+	}
+
+	if timeoutV, ok := syncInfo.TimeoutView(); ok &&
+		timeoutV > v && timeoutV < v+hotstuff.View(s.pipelinedViews) {
+		v = timeoutV
+	}
+
 	if v < s.currentView {
 		return
 	}
@@ -349,21 +379,21 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
 // This method is meant to be used instead of the exported UpdateHighQC internally
 // in this package when the qc has already been verified.
 func (s *Synchronizer) updateHighQC(qc hotstuff.QuorumCert) {
-
 	if qc.View() > s.highQC.View() {
 		s.highQC = qc
 		s.logger.Debug("HighQC updated")
 
 		// I believe the following can be removed, if leafBlock is updated elsewhere.
 		// However, placed here now for backwards compatibility
-		highQCBlock, ok := s.blockChain.Get(s.highQC.BlockHash())
-		if !ok {
-			s.logger.Info("updateHighQC: Could not find block referenced by new QC!")
-			return
-		}
-
-		if s.leafBlock.View() < s.highQC.View() || !s.blockChain.Extends(s.leafBlock, highQCBlock) {
-			s.leafBlock = highQCBlock
+		if s.leafBlock.View() < s.highQC.View() {
+			highQCBlock, ok := s.blockChain.Get(s.highQC.BlockHash())
+			if !ok {
+				s.logger.Info("updateHighQC: Could not find block referenced by new QC!")
+				return
+			}
+			if s.blockChain.Extends(s.leafBlock, highQCBlock) {
+				s.leafBlock = highQCBlock
+			}
 		}
 	}
 }
--- a/synchronizer/synchronizer.go
+++ b/synchronizer/synchronizer.go
@@ -349,20 +349,9 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
 // This method is meant to be used instead of the exported UpdateHighQC internally
 // in this package when the qc has already been verified.
 func (s *Synchronizer) updateHighQC(qc hotstuff.QuorumCert) {
-	newBlock, ok := s.blockChain.Get(qc.BlockHash())
-	if !ok {
-		s.logger.Info("updateHighQC: Could not find block referenced by new QC!")
-		return
-	}
 
-	oldBlock, ok := s.blockChain.Get(s.highQC.BlockHash())
-	if !ok {
-		s.logger.Panic("Block from the old highQC missing from chain")
-	}
-
-	if newBlock.View() > oldBlock.View() {
+	if qc.View() > s.highQC.View() {
 		s.highQC = qc
-		s.leafBlock = newBlock
 		s.logger.Debug("HighQC updated")
 	}
 }
--- a/synchronizer/synchronizer.go
+++ b/synchronizer/synchronizer.go
@@ -385,15 +385,14 @@ func (s *Synchronizer) updateHighQC(qc hotstuff.QuorumCert) {
 
 		// I believe the following can be removed, if leafBlock is updated elsewhere.
 		// However, placed here now for backwards compatibility
-		if s.leafBlock.View() < s.highQC.View() {
-			highQCBlock, ok := s.blockChain.Get(s.highQC.BlockHash())
-			if !ok {
-				s.logger.Info("updateHighQC: Could not find block referenced by new QC!")
-				return
-			}
-			if s.blockChain.Extends(s.leafBlock, highQCBlock) {
-				s.leafBlock = highQCBlock
-			}
+		highQCBlock, ok := s.blockChain.Get(s.highQC.BlockHash())
+		if !ok {
+			s.logger.Info("updateHighQC: Could not find block referenced by new QC!")
+			return
+		}
+
+		if s.leafBlock.View() < s.highQC.View() || !s.blockChain.Extends(s.leafBlock, highQCBlock) {
+			s.leafBlock = highQCBlock
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


synchronizer/synchronizer.go
```diff
@@ -315,6 +333,31 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
...
 	}
 
+	b, ok := syncInfo.Block()
+	if !ok {
+		b = s.leafBlock
+	}
+
+	highQCBlock, ok := s.blockChain.LocalGet(s.highQC.BlockHash())
+	if !ok {
+		s.logger.Error("AdvanceView: Could not find block referenced by new QC!")
```
meling, line 343: Is it safe to continue after this? Needless to say, if `highQCBlock` is nil, `Extends()` below will give a nil pointer panic.

synchronizer/synchronizer.go
```diff
@@ -315,6 +333,31 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
 		}
 	}
 
+	b, ok := syncInfo.Block()
```
meling, line 336: Perhaps this could be made into a separate method to reduce the method's lines of code and its cyclomatic complexity.

types.go
```diff
@@ -172,10 +172,13 @@ func (pc PartialCert) ToBytes() []byte {
...
+// This can also hola a block.
 type SyncInfo struct {
-	qc    *QuorumCert
-	tc    *TimeoutCert
-	aggQC *AggregateQC
+	qc         *QuorumCert
+	tc         *TimeoutCert
+	aggQC      *AggregateQC
+	block      *Block
+	timoutView View
```
meling, line 181: `s/timoutView/timeoutView/`

types.go
```diff
@@ -172,10 +172,13 @@ func (pc PartialCert) ToBytes() []byte {
 // Generally, if highQC.View > highTC.View, there is no need to include highTC in the SyncInfo.
 // However, if highQC.View < highTC.View, we should still include highQC.
 // This can also hold an AggregateQC for Fast-Hotstuff.
+// This can also hola a block.
```
meling, line 175: hold

synchronizer/synchronizer.go
```diff
@@ -349,21 +392,21 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
...
-		s.logger.Panic("Block from the old highQC missing from chain")
-	}
-
-	if newBlock.View() > oldBlock.View() {
+	if qc.View() > s.highQC.View() {
 		s.highQC = qc
-		s.leafBlock = newBlock
 		s.logger.Debug("HighQC updated")
+
+		// I believe the following can be removed, if leafBlock is updated elsewhere.
```
meling, line 399: Maybe we should add a TODO here or create an issue on the issue tracker to explain what should be done instead.

synchronizer/synchronizer.go
```diff
@@ -36,6 +36,9 @@ type Synchronizer struct {
 	duration ViewDuration
 	timer    *time.Timer
 
+	// number of cuncurrent views, 0 and 1 both give no concurrency.
```
meling, line 39: concurrent

synchronizer/synchronizer.go
```diff
@@ -192,6 +205,11 @@ func (s *Synchronizer) OnLocalTimeout() {
 	s.consensus.StopVoting(s.currentView)
 
 	s.configuration.Timeout(timeoutMsg)
+
+	if s.currentView < s.highQC.View()+hotstuff.View(s.pipelinedViews) {
```
meling, line 209: This construction:
```go
s.currentView < s.highQC.View()+hotstuff.View(s.pipelinedViews)
```
is used three times in this file. And could beneficially be done with a method:
```go
func (s Synchronizer) canAdvance(view hotstuff.View) bool {
  return view < s.highQC.View()+s.pipelinedViews
}
```
I think the compiler will most likely inline this code anyway, so shouldn't give any added overheads.

synchronizer/synchronizer.go
```diff
@@ -36,6 +36,9 @@ type Synchronizer struct {
 	duration ViewDuration
 	timer    *time.Timer
 
+	// number of cuncurrent views, 0 and 1 both give no concurrency.
+	pipelinedViews uint8
```
meling, line 40: You could make this be of the type `hotstuff.View`; then you wouldn't need to cast it all the time, and could add it directly to other views. But maybe it would be confusing?

meling (changes requested):
Looks good. I've requested some changes. Looks like you already refer to the number of views as the type `hotstuff.View`:
```go
	// PipelinedViews returns the number of concurrently executed views.
 	PipelinedViews() hotstuff.View
```
This makes it seem reasonable that the `Synchronizer` struct should also represent it as a `hotstuff.View` instead of a `uint8`.

synchronizer/synchronizer.go
leandernikolaus: Fixed in 600a0f7.

synchronizer/synchronizer.go
leandernikolaus: moved to updateLeafBlock

types.go
leandernikolaus: fixed in [7813a4e](https://github.com/relab/hotstuff/pull/78/commits/7813a4eb48467f0f2b55986620b6b5e0ee3abe60).

types.go
leandernikolaus: fixed in [7813a4e](https://github.com/relab/hotstuff/pull/78/commits/7813a4eb48467f0f2b55986620b6b5e0ee3abe60).

synchronizer/synchronizer.go
leandernikolaus: I removed the comment. I believe it may be important to only update the high QC with a block that actually exists.

synchronizer/synchronizer.go
leandernikolaus: @meling For me it seems a bit unlogical that you can add two views. 
Do you think, it would be better to keep `pipelinedViews` an `uint` and then define an add function on the view type?

leandernikolaus: Done in [eb95aa1](https://github.com/relab/hotstuff/pull/78/commits/eb95aa1b566321e76dec2296f05e3403fe58758e)

synchronizer/synchronizer.go
leandernikolaus: done in [eb95aa1](https://github.com/relab/hotstuff/pull/78/commits/eb95aa1b566321e76dec2296f05e3403fe58758e) but see comment above.

synchronizer/synchronizer.go
meling: > @meling For me it seems a bit unlogical that you can add two views.
> Do you think, it would be better to keep `pipelinedViews` an `uint` and then define an add function on the view type?

Using `uint` is fine, if we use a func to “hide” the type cast… it was anyway being casted before added to a view…

synchronizer/synchronizer.go
meling: Sorry, I wasn't clear enough in my message. I suggested moving all of the new code related to pipelining into a separate method, not just this, but it did help I guess.

synchronizer/synchronizer.go
```diff
@@ -339,31 +371,54 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
...
-		s.consensus.Propose(syncInfo)
+		s.consensus.Propose(syncInfo.WithQC(s.highQC))
 	} else if replica, ok := s.configuration.Replica(leader); ok {
 		replica.NewView(syncInfo)
 	}
 }
 
+// isInPipelineStretch checks wether the given view lies
+// less then the pipelinestretch from the highQC
+func (s *Synchronizer) isInPipelineStretch(v hotstuff.View) bool {
```
meling, line 382: What about just `inPipeline` or `isInPipeline` or `inPipelineRange`?

Typo in comment: wether -> whether. But perhaps this is more clear?:
```go
// inPipeline returns true if the given view is in pipeline range from highQC.
```

synchronizer/synchronizer.go
```diff
@@ -339,31 +371,54 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
...
+
+		if s.leafBlock.View() < s.highQC.View() || !s.blockChain.Extends(s.leafBlock, highQCBlock) {
+			s.leafBlock = highQCBlock
+		}
 	}
+}
 
-	oldBlock, ok := s.blockChain.Get(s.highQC.BlockHash())
+// updateLeafBlock attempts to update the leafBlock.
+// This method ensures, leafblock extends highQC.
```
meling, line 408: leaf block

meling: Also remove comma

synchronizer/synchronizer.go
meling: I've now reviewed the changes in more detail. I think it is fine to use `uint8` since it would essentially only need to cast it to `hotstuff.View` one additional place (in the `isInPipeline...()`), which would be fine since it wouldn't be part of code people read. So feel free to revert it to `uint8`.

synchronizer/synchronizer.go
```diff
@@ -315,6 +333,20 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
...
+	b, ok := syncInfo.Block()
+	if ok {
+		s.updateLeafBlock(b)
+	}
+
+	if s.isInPipelineStretch(s.leafBlock.View()) {
+		v = s.leafBlock.View()
+	}
+
+	if timeoutV, ok := syncInfo.TimeoutView(); ok &&
```
meling, line 345: Returning  `ok` is not necessary here. You can just do:
```go
	if timeoutV := syncInfo.TimeoutView(); timeoutV > v && s.isInPipelineStretch(timeoutV) {
```
Since `timeoutV == 0` will render the if-condition false, as with the current `ok` version.

synchronizer/synchronizer.go
```diff
@@ -315,6 +333,20 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
 		}
 	}
 
+	b, ok := syncInfo.Block()
```
meling, line 336: Not sure the `ok` return thing adds much value; isn't this just as clear?
```go
if b := syncInfo.Block(); b != nil {
	s.updateLeafBlock(b)
}
```

synchronizer/synchronizer.go
leandernikolaus: Fixed in [df3fc32](https://github.com/relab/hotstuff/pull/78/commits/df3fc324538afeeadc39d8adedcddb7d618f4ee8)

synchronizer/synchronizer.go
leandernikolaus: fixed in [df3fc32](https://github.com/relab/hotstuff/pull/78/commits/df3fc324538afeeadc39d8adedcddb7d618f4ee8)
```Go
// inPipeline checks if the given view is in pipeline range from highQC.
func (s *Synchronizer) inPipeline(v hotstuff.View) bool {
```

synchronizer/synchronizer.go
leandernikolaus: I changed it to `uint32` in [df3fc32](https://github.com/relab/hotstuff/pull/78/commits/df3fc324538afeeadc39d8adedcddb7d618f4ee8). 
This avoids a cast, since it is transferred by protobuf as `uint32`
```proto
 // Number of blocks proposed concurrently
  uint32 PipelinedViews = 22;
```

synchronizer/synchronizer.go
leandernikolaus: I guess that is the same for QC and TC. Check the following:
```go
// TC returns the timeout certificate, if present.
func (si SyncInfo) TC() (_ TimeoutCert, _ bool) {
	if si.tc != nil {
		return *si.tc, true
	}
	return
}

// AggQC returns the AggregateQC, if present.
func (si SyncInfo) AggQC() (_ AggregateQC, _ bool) {
	if si.aggQC != nil {
		return *si.aggQC, true
	}
	return
}

// Block returns the block, if present.
func (si SyncInfo) Block() (_ *Block, _ bool) {
	if si.block != nil {
		return si.block, true
	}
	return
}
```

synchronizer/synchronizer.go
leandernikolaus: fixed in [aae49db](https://github.com/relab/hotstuff/pull/78/commits/aae49db4ba31404475852512adf01459e04e32f4)
```go
// TimeoutView returns the timeout view, if present.
func (si SyncInfo) TimeoutView() (_ View) {
	return si.timeoutView
}
```

twins/network.go
```diff
@@ -130,6 +130,13 @@ func (n *Network) createTwinsNodes(nodes []NodeID, scenario Scenario, consensusN
 			return err
 		}
 
+		if pipelinedViews < 1 || consensusName != "chainedhotstuff" {
```
meling, line 133: This should not be done in the loop, and it is also more complex than necessary. Adding this at line 124 should suffice:
```go
	if pipelinedViews > 1 && consensusName != "chainedhotstuff" {
		return fmt.Errorf("pipelining currently only supported for chainedhotstuff")
	}
	if pipelinedViews < 1 {
		pipelinedViews = 1
	}
```

internal/mocks/synchronizer_mock.go
```diff
@@ -114,3 +114,17 @@ func (mr *MockSynchronizerMockRecorder) ViewContext() *gomock.Call {
...
+	ret := m.ctrl.Call(m, "PipelinedViews")
+	ret0, _ := ret[0].(hotstuff.View)
+	return ret0
+}
+
+// PipelinedViews indicates an expected call of PipelinedViews.
+func (mr *MockSynchronizerMockRecorder) PipelinedViews() *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "PipelinedViews", reflect.TypeOf((*MockSynchronizer)(nil).PipelinedViews))
+}
```
meling, line 130: Missing newline

twins/network.go
leandernikolaus: fixed in [fb04815](https://github.com/relab/hotstuff/pull/78/commits/fb04815f86cc71243e680d698634e0b055489b3b)
I also added a similar error in `internal/cli/run.go`
```go
func checkFlags() error {
	consensus := viper.GetString("consensus")
	pipelinedViews := viper.GetInt("pipelined-views")
	if pipelinedViews > 1 && consensus != "chainedhotstuff" {
		return errors.New("piplining only supported for chainedhotstuff consensus")
	}
	return nil
}
```

internal/mocks/synchronizer_mock.go
leandernikolaus: fixed in [fb04815](https://github.com/relab/hotstuff/pull/78/commits/fb04815f86cc71243e680d698634e0b055489b3b)

leandernikolaus:
In [7e59d24](https://github.com/relab/hotstuff/pull/78/commits/7e59d2464ddb7dba515b1b3d6792e726bc67101e) I revised the approach. Now the synchronizer has both a `currentView` and `nextView`. 
`currentView` is advanced on receiving a TC or QC, while `nextView` is advanced on receiving a block if it lies within the pipeline.
This makes the system run more stable since it no longer reacts to individual timeouts.
However, the consensus module now no longer proposes blocks only in the current view but needs to receive a view, in which it should propose.
This also required updates to Twins.
Based on my analysis and testing, without pipelining, `nextView` and `currentView` should always coincide.

meling:
@leandernikolaus I tried to merge master into this branch (locally) in an attempt to resolve the conflicts. However, the current synchronizer no longer has a `leafBlock` field and `LeafBlock()` method, it wasn't clear to me what needed to be done. This was removed in this [commit](https://github.com/relab/hotstuff/commit/ba90eb20079d5bff0b6d6a44c4fe482a8f10daa8). Could you try to resolve the conflicts, or let me know what needs to be done to get this back into a good shape. Or let me know if this no longer makes sense to merge.

PR #81, opened 2022-10-14
Reportexecutedcount

leandernikolaus:
This fixes a nondeterministically failing test in the pipeline.


internal/orchestration/worker.go
```diff
@@ -273,6 +274,7 @@ func (w *Worker) stopReplicas(req *orchestrationpb.StopReplicaRequest) (*orchest
 		}
 		r.Stop()
 		res.Hashes[id] = r.GetHash()
+		res.Counts[id] = r.GetCnt()
```
meling, line 277: Rename to `GetCmdCount()`

internal/orchestration/controller.go
```diff
@@ -342,14 +343,31 @@ func (e *Experiment) stopReplicas() error {
...
+				}
+			}
+		}
 	}
-	var cmp []byte
-	for _, hash := range hashes {
-		if cmp == nil {
-			cmp = hash
+	var count uint32
+	first := true
```
meling, line 360: Not sure what happened here, but in my browser, it looks like there is a newline after the `=` sign. But I don't see it in VSCode. Maybe it will go away with a new commit touching this file...

internal/orchestration/controller.go
```diff
@@ -342,14 +343,31 @@ func (e *Experiment) stopReplicas() error {
 		for id, hash := range res.GetHashes() {
 			hashes[id] = hash
 		}
+		for id, cnt := range res.GetCounts() {
```
meling, line 346: What are Lines 343-348 doing? Merging the final hashes and command counts from all replicas? Won't this overwrite the hash/count value for a given replica's `id`? 

My reading of the code here is that `worker.StopReplica` is called on all replicas, each of which will return the map of their hashes and counts for each of their counterparts. Presumably, these will have overlapping `id` fields and so would overwrite the same `id` entry. What am I missing here?

internal/orchestration/controller.go
```diff
@@ -342,14 +343,31 @@ func (e *Experiment) stopReplicas() error {
...
-	for _, hash := range hashes {
-		if cmp == nil {
-			cmp = hash
+	var count uint32
+	first := true
+	for _, cnt := range counts {
+		if cnt == 0 {
+			return fmt.Errorf("no request executed")
+		}
+		if first {
```
meling, line 365: Line 359 could be replaced with the following:
```go
var count uint32
if len(counts) > 0 {
   count = counts[0]
}
```
Avoiding the `first` bool and Lines 365-368.

internal/orchestration/controller.go
```diff
@@ -342,14 +343,31 @@ func (e *Experiment) stopReplicas() error {
...
 	}
-	var cmp []byte
-	for _, hash := range hashes {
-		if cmp == nil {
-			cmp = hash
+	var count uint32
+	first := true
+	for _, cnt := range counts {
+		if cnt == 0 {
+			return fmt.Errorf("no request executed")
```
meling, line 363: Perhaps it makes sense to include the replica `id` in the error message:
```go
return fmt.Errorf("no requests executed by replica %d", id)
```

replica/replica.go
```diff
@@ -170,3 +170,8 @@ func (srv *Replica) Close() {
 func (srv *Replica) GetHash() (b []byte) {
 	return srv.clientSrv.hash.Sum(b)
 }
+
+// GetCnt returns the count of all executed commands.
+func (srv *Replica) GetCnt() (c uint32) {
```
meling, line 175: `GetCmdCount()`

internal/orchestration/controller.go
```diff
@@ -342,14 +343,31 @@ func (e *Experiment) stopReplicas() error {
...
 		}
+		for id, cnt := range res.GetCounts() {
+			counts[id] = cnt
+		}
+	}
+	for id1, cnt1 := range counts {
+		for id2, cnt2 := range counts {
+			if id2 > id1 && cnt2 == cnt1 {
+				if !bytes.Equal(hashes[id1], hashes[id2]) {
+					return fmt.Errorf("hash mismatch for same count")
```
meling, line 354: Not sure if it is necessary or interesting, but would it make sense to include the mismatching IDs in the error message?

meling (changes requested):
There are a few nitpicks and one issue that is unclear to me; it looks like a bug. And adding some comments would ease the readability of the code (avoiding confusion between command ID and replica ID). I now see that the IDs are replica IDs, but it wasn't clear to me initially.

meling:
The tests look like they failed, but: *The logs for this run have expired and are no longer available.* Hence, it needs to be clarified what the problem was.

meling:
Hanish could benefit from this sanity check in his work.

@leandernikolaus Could you clarify what's going on in `internal/orchestration/controller.go` (see my [comment](https://github.com/relab/hotstuff/pull/81#discussion_r1103154195) above.

meling:
Closing as superseded by #88.

PR #87, opened 2023-02-14
Latency

hanish520:
Ability to identify the latency of the quorum certificate


internal/proto/hotstuffpb/hotstuff.proto
```diff
@@ -81,6 +85,12 @@ message QuorumCert {
   QuorumSignature Sig = 1;
   uint64 View = 2;
   bytes Hash = 3;
+  LatencyVector LatencyVector = 4;
```
meling, line 88: Looks like you could do this instead, and avoid the indirections to get the map via the `LatencyVector` message type:
```proto
map<uint32, uint64> LatencyVector =4;
```
This will also avoid encoding the extra information related to the `LatencyVector` message type, such as its position number etc.

That said, I think we could represent this vector much more compactly.

1. Do you really need 64 bits for the latency (assuming ms granularity)?
  - With 32 bits you can represent `2^32/(1000*60*60*24) = 49` days.
  - With 24 bits you can represent `2^24/(1000*60*60) = 4.66` hours.
2. Similarly, I don't think we will ever approach 32 bit `hotstuff.ID`s in the vector / configuration.

Some alternative encodings:
```proto
repeated uint32 LatencyVector = 4;
repeated uint64 LatencyVector = 4;
bytes LatencyVector = 4;
```

With the first one, you could use lower 24 bits for the latency value, and the upper 8 bits for the `hotstuff.ID`. For the second one, I guess splitting it into two `uint32` values for each is fine. And the last one is even more flexible, but is perhaps a bit more awkward to work with.

To test out / support the first option, you would only need to do something like this in `hotstuffpb/convert.go`:
```go
func QuorumCertFromProto(qc *QuorumCert) hotstuff.QuorumCert {
	var hash hotstuff.Hash
	copy(hash[:], qc.GetHash())
	latencyVector := make(map[uint32]uint64)
	for _, val := range qc.GetLatencies() {
		id := val & 0xff00_0000 >> 24
		latency := val & 0x00ff_ffff
		latencyVector[id] = uint64(latency)
	}
	return hotstuff.NewQuorumCert(
		hotstuff.ID(qc.Creator),
		QuorumSignatureFromProto(qc.GetSig()),
		hotstuff.View(qc.GetView()),
		hash,
		latencyVector,
	)
```
And similar on the sending side.

leaderrotation/roundrobin.go
```diff
@@ -30,5 +32,6 @@ func NewRoundRobin() modules.LeaderRotation {
 }
 
 func chooseRoundRobin(view hotstuff.View, numReplicas int) hotstuff.ID {
-	return hotstuff.ID(view%hotstuff.View(numReplicas) + 1)
+	id := uint64(math.Ceil(float64(view) / 2.0))
```
meling, line 35: I didn't understand this change. Is it related to the latency? 

Was to original version wrong/problematic? Maybe add an explainer comment here.

modules/modules.go
```diff
@@ -147,6 +147,12 @@ type LeaderRotation interface {
 	GetLeader(hotstuff.View) hotstuff.ID
 }
 
+// LatencyMeasurement implements the kauri protocol.
```
meling, line 150: Does it implement the kauri protocol?

crypto/crypto.go
```diff
@@ -55,7 +61,23 @@ func (c crypto) CreateQuorumCert(block *hotstuff.Block, signatures []hotstuff.Pa
 	if err != nil {
 		return hotstuff.QuorumCert{}, err
 	}
-	return hotstuff.NewQuorumCert(sig, block.View(), block.Hash()), nil
+	latencyVector := c.prepareLatencyVector(block, signatures)
+	return hotstuff.NewQuorumCert(c.opts.ID(), sig, block.View(), block.Hash(), latencyVector), nil
+}
+
+func (c crypto) prepareLatencyVector(block *hotstuff.Block, signatures []hotstuff.PartialCert) map[uint32]uint64 {
+	latencyVector := make(map[uint32]uint64)
```
meling, line 69: This is a general comment that applies in many places. Here you are allocating memory for the map and if you return early below, the map is empty. I think this may cause some unnecessary memory pressure and slow things down unnecessarily. I think it should be fine to return `nil` here (and other places). If you later try to iterate over a `nil` map, it will be as if it was an empty map.

You may need to add some `nil` guards in other places, but didn't check.

crypto/crypto.go
```diff
@@ -55,7 +61,23 @@ func (c crypto) CreateQuorumCert(block *hotstuff.Block, signatures []hotstuff.Pa
...
+	if !c.opts.IsLatencyCalculationEnabled() || c.opts.ID() != block.Proposer() {
+		return latencyVector
+	}
+
+	for _, sig := range signatures {
+		//timeDiff := sig.Time().UnixNano() - block.Time().UnixNano()
+		//timeDiff := sig.Time().UnixMilli() - block.Time().UnixMilli()
+		timeDiff := sig.Time().UnixMicro() - block.Time().UnixMicro()
```
meling, line 77: Suggest to use std lib methods for this subtraction:
```go
td := sig.Time().Sub(block.Time())
// if you need to take the absolute value, use:
td = td.Abs()
// to truncate to millisecond granularity
timeDiff := td.Truncate(time.Millisecond)
```

crypto/crypto.go
```diff
@@ -55,7 +61,23 @@ func (c crypto) CreateQuorumCert(block *hotstuff.Block, signatures []hotstuff.Pa
...
+		latencyVector[uint32(sig.Signer())] = uint64(timeDiff)
```
meling, line 78: Looks like this is the place to encode the latency into 32 or 24 bits and encode the signer with 32 or 8 bits, depending on the choice of proto representation for the latency vector. See other comment.

meling (changes requested):
Didn't get time to study the `latency/latency.go` file yet. Other than that I've requested some changes.

Some general comments are: 
- Be careful with allocation of maps and similar; they are convenient to use, but when used in the hot path, they can lead to perf bottlenecks.
- I suggest to use a more compact vector representation also in the code instead of a map; I haven't thought carefully about it, but if you look at Gonum's matrix representation they use a [single slice and row, col, and a stride parameter](https://github.com/gonum/gonum/blob/v0.12.0/blas/blas32/blas32.go#L37). This suggestion is independent of the representation used in the proto format, which is probably quite important for the message sizes.

latency/latency.go
```diff
@@ -0,0 +1,115 @@
...
+	for id, newValue := range latencyVector {
+		stats := l.latencyStats[event.Creator][hotstuff.ID(id)]
+		if stats == nil {
+			return
+		}
+		l.ComputeStats(stats, float64(newValue))
+	}
+}
+
+func (l *latencyMeasurement) GetLatencyMatrix() map[hotstuff.ID]map[hotstuff.ID]uint64 {
```
meling, line 81: I suggest to not expose the matrix as a `map[hotstuff.ID]map[hotstuff.ID]uint64` but instead define a new type that has access methods:

```go
type LatencyMatrix map[hotstuff.ID]map[hotstuff.ID]uint64

// Latency returns the latency between replicas a and b.
func (lm *LatencyMatrix) Latency(a, b hotstuff.ID) time.Duration {}

// LatencyVector returns the latency vector for replica a.
func (lm *LatencyMatrix) LatencyVector(a hotstuff.ID) []time.Duration {}
```

This way you avoid exposing the internal datastructure of the matrix, enabling future optimization without impacting the code using the matrix.

meling:
Ping @hanish520. Could you please review my suggestions so that we can get this merged soon.

hanish520:
> Ping @hanish520. Could you please review my suggestions so that we can get this merged soon.

Hein, i will look into it this weekend, we can do a live review next week when you are here.

internal/orchestration/controller.go
```diff
@@ -270,7 +279,7 @@ func (e *Experiment) assignReplicasAndClients() (err error) {
 			replicaOpts := proto.Clone(e.ReplicaOpts).(*orchestrationpb.ReplicaOpts)
 			replicaOpts.ID = uint32(nextReplicaID)
 			replicaOpts.ByzantineStrategy = byzantineStrategy
-
+			e.replicaLocationInfo[nextReplicaID] = location
```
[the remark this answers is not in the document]
hanish520: Hein, we need this map in the experiment struct, by the moment we assign the opts here, we don't have the complete location assignment map of the replicas. once the allocation is complete we are copying it to the opts structure to be used by the replicas for initialization.

hanish520:
@meling I reviewed the comments, and now I realized that optiblog latency recording code changes were entered into this PR. I may abandon this PR and create a new PR with changes required only for inducing latency in the framework.

PR #89, opened 2023-02-21
Kauri module implementation

hanish520:
kauri implementation first draft


backend/server.go
```diff
@@ -134,16 +135,18 @@ type serviceImpl struct {
...
 	proposeMsg := hotstuffpb.ProposalFromProto(proposal)
-	proposeMsg.ID = id
-
+	if !impl.srv.opts.ShouldUseKauri() {
+		id, err := GetPeerIDFromContext(ctx, impl.srv.configuration)
+		if err != nil {
+			impl.srv.logger.Infof("Failed to get client ID: %v", err)
+			return
+		}
+		proposal.Block.Proposer = uint32(id)
```
meling, line 145: Given that `proposal` isn't used later in this method, will this have an effect?
Or is this one still necessary: `proposeMsg := hotstuffpb.ProposalFromProto(proposal)`?

meling: It could perhaps be helpful to add a comment to explain why Kauri can use the ID from `proposal.Block.Proposer` while non-Kauri gets the ID from context.

consensus/consensus.go
```diff
@@ -153,8 +153,9 @@ func (cs *consensusBase) Propose(cert hotstuff.SyncInfo) {
 	}
 
 	cs.blockChain.Store(proposal.Block)
-
-	cs.configuration.Propose(proposal)
+	if cs.kauri == nil {
+		cs.configuration.Propose(proposal)
```
meling, line 157: Would it be useful to have a comment explaining that Kauri won't propose here, but elsewhere?

internal/proto/kauripb/kauri.proto
```diff
@@ -0,0 +1,23 @@
...
+  rpc SendContribution(Contribution) returns (google.protobuf.Empty) {
+    option (gorums.unicast) = true;
+  }
+}
+
+message Contribution {
+  uint32 ID = 1;
+  hotstuffpb.QuorumSignature Signature = 2;
+  uint64 View = 3;
+}
```
meling, line 23: newline

kauri/kauri.go
```diff
@@ -0,0 +1,320 @@
+// Package kauri contains the implementation of the kauri protocol
```
meling, line 1: ... the Kauri protocol.

kauri/kauri.go
```diff
@@ -0,0 +1,320 @@
...
+)
+
+func init() {
+	modules.RegisterModule("kauri", New)
+}
+
+// TreeType determines the type of tree to be built, for future extension
+type TreeType int
+
+// NoFaultTree is a tree without any faulty replicas in the tree, which is currently supported.
```
meling, line 30: These comments should be moved inside the const block above each of the constants so that the docs are generated correctly.

kauri/kauri.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+func (k *Kauri) initializeConfiguration() {
+	kauriCfg := kauripb.ConfigurationFromRaw(k.configuration.GetRawConfiguration(), nil)
+	for _, n := range kauriCfg.Nodes() {
+		k.nodes[hotstuff.ID(n.ID())] = n
+	}
+	k.tree = CreateTree(k.configuration.Len(), k.opts.ID())
+
+	// pIDs := make(map[hotstuff.ID]int)
```
meling, line 98: Still need this commented code?

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+type FaultFreeTree struct {
+	ID                  hotstuff.ID
+	ConfigurationLength int
+	height              int
+	idToPosMapping      map[hotstuff.ID]int
+	posToIDMapping      map[int]hotstuff.ID
+}
+
+// CreateTree Creates the tree configuration, currently only fault free tree configuration is supported.
+func CreateTree(configurationLength int, myID hotstuff.ID) TreeConfiguration {
```
meling, line 31: This is the interface anti-pattern in Go that I mentioned a while back. Generally, we avoid returning an interface type and defining interfaces before they are needed.

Suggest to rename this function to `newFaultFreeTree` and return `*faultFreeTree` instead of the interface. If you later add a new tree type, you can create another construction function that returns the other tree type.

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+type TreeConfiguration interface {
+	InitializeWithPIDs(ids map[hotstuff.ID]int)
+	GetHeight() int
+	GetChildren() []hotstuff.ID
+	GetSubTreeNodes() []hotstuff.ID
+	GetParent() (hotstuff.ID, bool)
+}
+
+// FaultFreeTree implements a fault free tree configuration.
+type FaultFreeTree struct {
```
meling, line 22: Consider whether or not all these types and methods and such needs to be exported... I haven't checked all methods etc, but looks like most of methods and types are internal to the `kauri` package and should not be exported.

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// CreateTree Creates the tree configuration, currently only fault free tree configuration is supported.
+func CreateTree(configurationLength int, myID hotstuff.ID) TreeConfiguration {
+
+	if configurationLength <= 0 {
+		return nil
+	}
+	temp := configurationLength
+	temp = temp - 1 //root
```
meling, line 37:
```go
temp := configurationLength - 1 // root
```
Avoids Line 36.

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+	height := 1
+	for i := 1; temp > 0; i++ {
```
meling, line 39: Maybe this could be converted to a helper function:
```go
func height(maxChild float64, treeNodes int) int {
	remainingNodes := treeNodes - 1 // root
	height := 1
	for i := 1; remainingNodes > 0; i++ {
		remainingNodes -= int(math.Pow(maxChild, float64(i)))
		height++
	}
	return height
}
```
And you could add this test:
```go
func TestTreeHeight(t *testing.T) {
	const MaxChild = 2
	tests := []struct {
		configurationLength int
		expectedHeight      int
	}{
		{1, 1},
		{2, 2},
		{3, 2},
		{4, 3},
		{5, 3},
		{6, 3},
		{7, 3},
		{8, 4},
		{9, 4},
		{10, 4},
		{11, 4},
		{12, 4},
		{13, 4},
		{14, 4},
		{15, 4},
		{16, 5},
		{17, 5},
	}
	for _, test := range tests {
		actualHeight := height(MaxChild, test.configurationLength)
		if actualHeight != test.expectedHeight {
			t.Errorf("height(%d) = %d, expected %d", test.configurationLength, actualHeight, test.expectedHeight)
		}
	}
}
```

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+// FaultFreeTree implements a fault free tree configuration.
+type FaultFreeTree struct {
+	ID                  hotstuff.ID
+	ConfigurationLength int
+	height              int
+	idToPosMapping      map[hotstuff.ID]int
+	posToIDMapping      map[int]hotstuff.ID
+}
+
+// CreateTree Creates the tree configuration, currently only fault free tree configuration is supported.
```
meling, line 30: Renaming this function to `newFaultFreeTree` this doc comment can be replaced with:
*newFaultFreeTree creates a fault free tree configuration.*

There is no need to say that it is the only supported tree type.

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	return t.posToIDMapping[(myPos-1)/MaxChild], true
+}
+
+// GetChildren returns the children of the replicas, if any.
+func (t *FaultFreeTree) GetChildren() []hotstuff.ID {
+	return t.GetChildrenOfNode(t.ID)
+}
+
+func (t *FaultFreeTree) isWithInIndex(position int) bool {
```
meling, line 73: Is there a better name for this? Could this be simply `contains(index int) bool`?

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+// InitializeWithPIDs uses the map to initialize the position of replicas.
+func (t *FaultFreeTree) InitializeWithPIDs(ids map[hotstuff.ID]int) {
+	t.idToPosMapping = ids
+	t.posToIDMapping = make(map[int]hotstuff.ID)
+	for ID, index := range ids {
+		t.posToIDMapping[index] = ID
+	}
+}
+
+// GetParent fetches the ID of the parent, if root, returns itself.
```
meling, line 59: In Go we typically don't use `Get` for getters... You can simply call this `Parent()`.

Also, the doc comment should maybe say something like:
```
Parent returns the tree node's parent ID and true.
Otherwise, if the tree node is the root, it returns its own ID and false.
```

kauri/kauri.go
```diff
@@ -0,0 +1,320 @@
...
+	totalNodes := k.configuration.Len()
+	ids := make([]hotstuff.ID, 0, totalNodes)
+	for id := range k.configuration.Replicas() {
+		ids = append(ids, id)
+	}
+	sort.Slice(ids, func(i, j int) bool { return ids[i] < ids[j] })
+	seed := k.opts.SharedRandomSeed() + int64(binary.LittleEndian.Uint64(hash[:]))
+	//Shuffle the list of IDs using the shared random seed + the first 8 bytes of the hash.
+	rnd := rand.New(rand.NewSource(seed))
+	rnd.Shuffle(len(ids), reflect.Swapper(ids))
```
meling, line 305: Should avoid reflect here. I think something like this will work:
```go
func(i, j int) { ids[i], ids[j] = ids[j], ids[i] })
```

kauri/kauri.go
```diff
@@ -0,0 +1,320 @@
...
+		k.SendContributionToParent()
+		k.isAggregationSent = true
+	}
+}
+
+// SendContributionToParent sends contribution to the parent node.
+func (k *Kauri) SendContributionToParent() {
+	parent, ok := k.tree.GetParent()
+	if ok {
+		node, isPresent := k.nodes[parent]
```
meling, line 188: Consider moving the `nodes` map to the tree and let `tree.Parent()` return the parent node directly instead of the `hotstuff.ID`.

kauri/kauri.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	for _, id := range a {
+		if !c.Contains(id) {
+			return false
+		}
+	}
+	return true
+}
+
+func (k *Kauri) randomizeIDS(hash hotstuff.Hash, leaderID hotstuff.ID) map[hotstuff.ID]int {
```
meling, line 293: Suggest to rename this to `shuffleTreeNodes` and move it into the tree. And instead of passing in the `hash` you could move the `seed` calculation out from this method... so that the signature becomes:
```go
func (t *faultFreeTree) shuffleTreeNodes(seed int64, leaderID hotstuff.ID)
```
This way you can update the internals of the tree directly avoiding the passing around of maps from this method to `InitializeWithPIDs` (which could be removed; its functionality can be done in the constructor and in the above method.)

kauri/kauri.go
```diff
@@ -0,0 +1,320 @@
...
+
+func (k *Kauri) postInit() {
+	k.logger.Info("Handel: Initializing")
+	kauripb.RegisterKauriServer(k.server.GetGorumsServer(), serviceImpl{k})
+	k.initializeConfiguration()
+}
+
+func (k *Kauri) initializeConfiguration() {
+	kauriCfg := kauripb.ConfigurationFromRaw(k.configuration.GetRawConfiguration(), nil)
+	for _, n := range kauriCfg.Nodes() {
```
meling, line 93: Seems better to move this into the tree constructor like this:
```go
func newFaultFreeTree(cfg kauripb.Configuration, myID hotstuff.ID) *FaultFreeTree {
	nodes := make(map[hotstuff.ID]*kauripb.Node)
	idMappings := make(map[hotstuff.ID]int)
	posToIDMapping := make(map[int]hotstuff.ID)
	for i, n := range cfg.Nodes() {
		nodes[hotstuff.ID(n.ID())] = n
		idMappings[hotstuff.ID(i+1)] = i
		posToIDMapping[i] = hotstuff.ID(i + 1)
	}
	// some code removed for brevity
	return &FaultFreeTree{
		ID:                  myID,
		ConfigurationLength: len(nodes),
		height:              height,
		nodes:               nodes,
		idToPosMapping:      idMappings,
		posToIDMapping:      posToIDMapping,
	}
}
```

Note: I don't think you actually need all three maps, but I didn't check. 
Also, it seems the `hotstuff.ID` generation in the above code could be using only a single one like this:
```go
	id := hotstuff.ID(n.ID())
	nodes[id] = n
	idMappings[id] = i
	posToIDMapping[i] = id
```

kauri/kauri.go
```diff
@@ -0,0 +1,320 @@
...
+		k.nodes[hotstuff.ID(n.ID())] = n
+	}
+	k.tree = CreateTree(k.configuration.Len(), k.opts.ID())
+
+	// pIDs := make(map[hotstuff.ID]int)
+	// for id := range k.configuration.ActiveReplicas() {
+	// 	pIDs[id] = index
+	// 	index++
+	// }
+	idMappings := make(map[hotstuff.ID]int)
```
meling, line 103: I believe this code is redundant with the suggested constructor code above.

kauri/kauri.go
```diff
@@ -0,0 +1,320 @@
...
+func (k *Kauri) Begin(pc hotstuff.PartialCert, p hotstuff.ProposeMsg) {
+	if !k.initDone {
+		k.eventLoop.DelayUntil(backend.ConnectedEvent{}, func() { k.Begin(pc, p) })
+		return
+	}
+	k.reset()
+	k.blockHash = pc.BlockHash()
+	k.currentView = p.Block.View()
+	k.aggregatedContribution = pc.Signature()
+	ids := k.randomizeIDS(k.blockHash, k.leaderRotation.GetLeader(k.currentView))
```
meling, line 122: This can be moved to the tree and merged with `InitializeWithPIDs()` as outlined in the [comment below](https://github.com/relab/hotstuff/pull/89/files#r1135014800).

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/relab/hotstuff"
+)
+
+// MaxChild determines the branching factor of the tree.
+const MaxChild = 2
+
+// TreeConfiguration is an abstraction for a tree communication model.
+type TreeConfiguration interface {
+	InitializeWithPIDs(ids map[hotstuff.ID]int)
+	GetHeight() int
```
meling, line 15: We avoid `Get` in Go.

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+	GetChildren() []hotstuff.ID
+	GetSubTreeNodes() []hotstuff.ID
+	GetParent() (hotstuff.ID, bool)
+}
+
+// FaultFreeTree implements a fault free tree configuration.
+type FaultFreeTree struct {
+	ID                  hotstuff.ID
```
meling, line 23: If I understand correctly this `ID` field is the current/local tree node. If so, perhaps we should use a more clear name.

meling: Also, maybe it shouldn't be exported.

kauri/kauri.go
meling: This said, I think a more idiomatic design would be to pass the `TreeConfiguration` as an argument to the `Karui` module. But I don't know how to do that with the module system.

kauri/tree.go
```diff
@@ -0,0 +1,170 @@
...
+	return position < t.ConfigurationLength
+}
+
+// GetGrandParent returns grand parent of a replica, if not possible, return highest parent and false.
+func (t *FaultFreeTree) GetGrandParent() (hotstuff.ID, bool) {
+	parent, ok := t.GetParent()
+	if !ok {
+		return parent, false
+	}
+	parentPos := t.idToPosMapping[parent]
```
meling, line 83:
```go
parentPos := t.idToPosMapping[parent] - 1
```
would avoid the two `parentPos-1` on Lines 84 and 87.

replica/replica.go
```diff
@@ -109,6 +109,7 @@ func New(conf Config, builder modules.Builder) (replica *Replica) {
 		srv.clientSrv.cmdCache,
 		srv.clientSrv.cmdCache,
 	)
+	//builder.Options().SetShouldVerifyVotesSync()
```
meling, line 112: Should this be commented as is, or enabled or removed?

meling (changes requested):
I didn't comment on all methods in FaultFreeTree and Kauri, but several of the methods are non-trivial to review. That is, there should be tests for these methods that operate with different indexes and boundaries that are easy to get wrong.

PR #96, opened 2023-04-14
Latency simulation support

hanish520:
This PR adds support for simulating latencies between replicas-replicas and clients-replicas.


internal/cli/run.go
```diff
@@ -205,6 +210,21 @@ func runController() {
 	checkf("failed to close ssh connections: %v", err)
 }
 
+func checkHostConfigLocations(location string) error {
```
meling, line 213: `s/checkHostConfigLocations/checkHost/` (since the first argument's name is `location` and is called with `cfg.Location`)

internal/cli/run.go
```diff
@@ -183,6 +183,11 @@ func runController() {
 
 	for _, cfg := range hostConfigs {
 		experiment.HostConfigs[cfg.Name] = cfg
+		if cfg.Location == "" {
+			cfg.Location = hotstuff.DefaultLocation
+		}
+		err := checkHostConfigLocations(cfg.Location)
+		checkf("invalid location configuration for host %s: %v", cfg.Name, err)
```
meling, line 190: Seem like this is a bit redundant: 
- the location is printed twice
- the error message doesn't add much value since the returned error already say `unknown location: %s`

Suggest to change this one to:
```
checkf("invalid configuration: %v", err)
```

backend/options.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"time"
+
+	"github.com/relab/gorums"
+	"github.com/relab/hotstuff"
+)
+
+type backendOptions struct {
+	location          string
+	locationInfo      map[uint32]string
```
meling, line 12: This could be `map[hotstuff.ID]string` and similarly on Line 21. Then you could avoid the type cast on Line 23. May also require changes elsewhere, e.g., at `WithLatencyInfo` call sites.

meling (changes requested):
Mostly minor changes requested, but please also see if you can fix the test/lint issues below at the bottom.

meling (approved):

PR #98, opened 2023-05-11
Fuzz test convert and twins scenario

abjeni:
This is a new testing tool using google/gofuzz to create a random hotstuff.pb.go message, and uses functions in convert.go to create a message that will be used in a twins scenario test by replacing it with the first message sent from a node in the scenario test. Crashes accumulates over a series of tests with different random messages and gets output in an organized list after the test is run 1000 times.

made by Asbjørn Salhus and Magnus Brandsegg

cd fuzz
go test -run TestFuzz


meling:
Not ready to merge this yet or it will be mergeable in the future..., but I've cleaned up some of the code, in particular related to printing protobuf messages as Go struct literals and replaced some of the confusing interface logic.

PR #102, opened 2023-07-26
modules/registry: fix ListModules with interface types

### LINKED ISSUE #101: --list-modules CLI flag produces segfault
The `--list-modules` CLI flag produces a segfault:

```
› make && ./hotstuff --list-modules
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xe8 pc=0x70ca1c]

goroutine 1 [running]:
github.com/relab/hotstuff/modules.ListModules()
	/home/michael/dev/go/hotstuff/modules/registry.go:89 +0x1dc
github.com/relab/hotstuff/internal/cli.glob..func1(0x11be760?, {0xc1f2a6?, 0x1?, 0x1?})
	/home/michael/dev/go/hotstuff/internal/cli/root.go:37 +0x34
github.com/spf13/cobra.(*Command).execute(0x11be760, {0xc0001a8010, 0x1, 0x1})
	/home/michael/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:856 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0x11be760)
	/home/michael/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:974 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
	/home/michael/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:902
github.com/relab/hotstuff/internal/cli.Execute()
	/home/michael/dev/go/hotstuff/internal/cli/root.go:52 +0x25
main.main()
	/home/michael/dev/go/hotstuff/cmd/hotstuff/main.go:7 +0x17
```
Following the stack trace shows it must be due to the `t` variable, a key of the `byInterface` map, being `<nil>`: https://github.com/relab/hotstuff/blob/master/modules/registry.go#L89

The queried datastructure (`byInterface`) is only filled within here, where the offending map key is called `moduleType`: https://github.com/relab/hotstuff/blob/master/modules/registry.go#L31-L34

`moduleType` is set in the same function as `moduleType := reflect.TypeOf(t)`. Here, `var t T` is (the zero value of) a variable of type `T`. `T` is seemingly the (generic) return type of the constructor of the module which is being registered.

Which brings us to the likely cause. The return type of at least some of the registered modules' constructors is an interface, e.g here: https://github.com/relab/hotstuff/blob/master/consensus/byzantine/byzantine.go#L11

Thus `var t T` has the zero value of an interface, which is `<nil>`. `moduleType := reflect.TypeOf(t)` [returns `nil` as `t` is a `nil` interface](https://pkg.go.dev/reflect#TypeOf). This ends up in the `byInterface` map. And eventually it blows up when one tries to access it as something of type `reflect.Type`.

I don't understand why you seemingly keep modules grouped by their constructor's return value in the `byInterface` data structure, and why you want to print them that way in the `--list-modules` call, so am hesitant to recommend a fix.

johningve:
This was broken since fc203fc9

The problem is that RegisterModule looks at the zero value of a module's type. However, this does not work for interface types. A workaround is to get the type of a pointer to the interface and then access its element type.

Fixes #101

### DIFF
--- a/modules/registry.go
+++ b/modules/registry.go
@@ -14,10 +14,11 @@ var (
 
 // RegisterModule registers a constructor for a module implementation with the specified name.
 // For example:
-//  RegisterModule("chainedhotstuff", func() consensus.Rules { return chainedhotstuff.New() })
+//
+//	RegisterModule("chainedhotstuff", func() consensus.Rules { return chainedhotstuff.New() })
 func RegisterModule[T any](name string, constructor func() T) {
 	var t T
-	moduleType := reflect.TypeOf(t)
+	moduleType := reflect.TypeOf(&t).Elem()
 
 	registryMut.Lock()
 	defer registryMut.Unlock()
@@ -39,7 +40,8 @@ func RegisterModule[T any](name string, constructor func() T) {
 // GetModule constructs a new instance of the module with the specified name.
 // GetModule returns true if the module is found, false otherwise.
 // For example:
-//  rules, ok := GetModule[consensus.Rules]("chainedhotstuff")
+//
+//	rules, ok := GetModule[consensus.Rules]("chainedhotstuff")
 func GetModule[T any](name string) (out T, ok bool) {
 	targetType := reflect.TypeOf(out)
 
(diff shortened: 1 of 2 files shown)


meling (approved):
Looks good to me!

PR #105, opened 2023-08-12
Fix synchronizer timer

johningve:
The timer used by the synchronizer used to be resettable, but after
https://github.com/relab/hotstuff/commit/7776a29f8668f3f20b221b57164ab6597c3c2fc6, the closure passed to the timer should only be used once
because it stores the current view. This was done so that we can avoid
the need for any synchronization of access to the current view in the
synchronizer. However, I failed to realize that this would make the
timer only usable once. Thus, resetting it is not possible as the view
that is stored in the timer's closure would not change upon reset.

This PR wraps the timer in a `oneShotTimer` struct that is intended to
prevent the timer from being reused. Instead, the methods
`startTimeoutTimer` and `stopTimeoutTimer` should be used instead of
using the timer directly. These functions ensure that each timer is
unique and thus a unique closure is used each time, ensuring that we get
the correct view and avoid the need for synchronization code.


twins/network.go
```diff
@@ -143,7 +143,7 @@ func (n *Network) createTwinsNodes(nodes []NodeID, _ Scenario, consensusName str
 			consensus.New(consensusModule),
 			consensus.NewVotingMachine(),
 			crypto.NewCache(ecdsa.New(), 100),
-			synchronizer.New(FixedTimeout(0)),
+			synchronizer.New(FixedTimeout(1*time.Millisecond)),
```
meling, line 146: We should consider if it would be enough to use e.g., `100 * time.Microsecond` to avoid that twins tests take longer time. This should be tested and maybe "measured" since I'm not sure this will have an impact, one way or the other.

meling (approved):
This looks good. I added a comment in twins/network.go... but I'm okay to merge it as is. We can always tweak the timeout later if we think it can save us some test execution time.

twins/network.go
hanish520: I tried with different values, it looks like 1 millisecond is more stable compared to lower durations.

PR #112, opened 2024-02-25
feat(eddsa): Added EDDSA crypto module

hanish520:
Added EDDSA(25519 curve) crypto module to the framework.
 The current design for eddsa module is messy, this PR is a starting point to take this activity.
Need to discuss with others.


JusticeGu:
as run ./hotstuff run is show "panic: interface conversion: hotstuff.QuorumSignature is crypto.MultiSignature[github.com/relab/hotstuff/crypto.Signature], not crypto.MultiSignature[*github.com/relab/hotstuff/crypto/ecdsa.Signature]
"

crypto/multi_signature.go
```diff
@@ -0,0 +1,89 @@
+package crypto
+
+import (
+	"reflect"
+
+	"github.com/relab/hotstuff"
+	"golang.org/x/exp/slices"
```
meling, line 7: This can be replaced with just `"slices"` I think. Or it may need the dependency update to get go1.22, but not totally sure. At least if we merge #118 it should work without this `x/exp/slices` package.

crypto/multi_signature.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// Signature is the individual component in MultiSignature
+type Signature interface {
+	Signer() hotstuff.ID
+	ToBytes() []byte
+}
+
+// MultiSignature is a set of (partial) signatures.
+type MultiSignature[T Signature] map[hotstuff.ID]T
```
meling, line 17: Suggest to rename `MultiSignature` to just `Multi` to avoid the stutter. We almost always use it together with something like this `Multi[*Signature]` and so it is still clear that it is a multi signature.

crypto/multi_signature.go
```diff
@@ -0,0 +1,89 @@
...
+
+// RestoreMultiSignature should only be used to restore an existing threshold signature from a set of signatures.
+func RestoreMultiSignature[T Signature](signatures []T) MultiSignature[T] {
```
meling, line 20: Should rename this too: `RestoreMultiSignature` to `RestoreMulti` or just `Restore`

crypto/keygen/keygen.go
```diff
@@ -316,14 +342,22 @@ func GenerateKeyChain(id hotstuff.ID, validFor []string, crypto string, ca *x509
...
 	case "ecdsa":
 		privateKey = ecdsaKey
+		publicKey = privateKey.Public()
 	case "bls12":
 		privateKey, err = bls12.GeneratePrivateKey()
 		if err != nil {
 			return KeyChain{}, fmt.Errorf("failed to generate bls12-381 private key: %w", err)
 		}
+		publicKey = privateKey.Public()
+	case "eddsa":
```
meling, line 356: Could we replace these case strings with their corresponding `{ecdsa, bls12, eddsa}.PublicKeyFileType`

meling (changes requested):
A few small adjustments should do it.

meling (approved):

PR #123, opened 2024-03-10
Fix eddsa private key parsing bug

hanish520:
The eddsa private key is encoded as PEM(PKCS#8) format. While decoding the private key a bug was introduced to generate a new key from the encoded bytes.
Fixed the issue by decoding the private key with same format.

### DIFF
--- a/crypto/keygen/keygen.go
+++ b/crypto/keygen/keygen.go
@@ -233,7 +233,17 @@ func ParsePrivateKey(buf []byte) (key hotstuff.PrivateKey, err error) {
 	case ecdsacrypto.PrivateKeyFileType:
 		key, err = x509.ParseECPrivateKey(b.Bytes)
 	case eddsa.PrivateKeyFileType:
-		key = ed25519.NewKeyFromSeed(b.Bytes[:32])
+		//key = ed25519.NewKeyFromSeed(b.Bytes[:32])
+		var genericKey any
+		var ok bool
+		genericKey, err = x509.ParsePKCS8PrivateKey(b.Bytes)
+		if err != nil {
+			return nil, err
+		}
+		key, ok = genericKey.(ed25519.PrivateKey)
+		if !ok {
+			return nil, fmt.Errorf("failed to parse key")
+		}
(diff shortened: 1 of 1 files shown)


crypto/keygen/keygen.go
```diff
@@ -233,7 +233,17 @@ func ParsePrivateKey(buf []byte) (key hotstuff.PrivateKey, err error) {
 	case ecdsacrypto.PrivateKeyFileType:
 		key, err = x509.ParseECPrivateKey(b.Bytes)
 	case eddsa.PrivateKeyFileType:
-		key = ed25519.NewKeyFromSeed(b.Bytes[:32])
+		//key = ed25519.NewKeyFromSeed(b.Bytes[:32])
```
meling, line 236: Delete

crypto/keygen/keygen.go
```diff
@@ -233,7 +233,17 @@ func ParsePrivateKey(buf []byte) (key hotstuff.PrivateKey, err error) {
...
+		var genericKey any
+		var ok bool
+		genericKey, err = x509.ParsePKCS8PrivateKey(b.Bytes)
```
meling, line 239: You can do:
```go
genericKey, err :=
```
To avoid the `var genericKey any` line above.

crypto/keygen/keygen.go
```diff
@@ -233,7 +233,17 @@ func ParsePrivateKey(buf []byte) (key hotstuff.PrivateKey, err error) {
 	case ecdsacrypto.PrivateKeyFileType:
 		key, err = x509.ParseECPrivateKey(b.Bytes)
 	case eddsa.PrivateKeyFileType:
-		key = ed25519.NewKeyFromSeed(b.Bytes[:32])
+		//key = ed25519.NewKeyFromSeed(b.Bytes[:32])
+		var genericKey any
+		var ok bool
```
meling, line 238: Move `ok` closer to its usage point.

meling (changes requested):
Minor changes

meling (approved):

PR #137, opened 2024-12-20
Replace old location and latency management with cleaner and easier to use design

meling:
This PR makes the location and latencies management easier to manage and use for ourselves.
This is in preparation for an improved configuration file design and orchestration design.
This work will arrive in a follow up PR.

- **feat: revised latencygen to generate latency matrix from csv file**
- **refactor(latencygen): 2D slice for latencies and slice for cities**
- **feat(latencies): added latency API functions to lookup latencies**
- **fix(latencygen): add city index and name in per-line comment**
- **refactor(latencygen): rename cities slice to locations**
- **feat(latencies): add Latencies(loc) and LocationName()**
- **feat(latencies): add LatenciesFrom(loc []string) LatencyMatrix**
- **refactor(latencies): removed Locations() function**
- **refactor(latencies): removed Latency(i, j) function**
- **refactor(latencies): removed LatencyID(a, b hotstuff.ID) function**
- **refactor(latencies): removed Latencies(location) function**
- **refactor(latencies): removed LocationName(index) function**
- **refactor(latencies): renamed LatencyCity and LatencyID to Latency**
- **refactor(latency): rename package to latency and variable renaming**
- **refactor(backend): uses new streamlined latency.Matrix to add delay**


backend/server.go
```diff
@@ -63,15 +62,15 @@ func NewServer(opts ...ServerOptions) *Server {
 	return srv
 }
 
-func (srv *Server) induceLatency(sender hotstuff.ID) {
-	if srv.location == hotstuff.DefaultLocation {
+// addNetworkDelay adds latency between this server and the sender based on
+// the latency between the two locations according to the latency matrix.
+func (srv *Server) addNetworkDelay(sender hotstuff.ID) {
```
AlanRostem, line 67: Suggestion: replace all usage of this method by the `Delay()` method of `latency.Matrix{}` since the logic is identical.

Ideally, keep this method and only replace the `time.Sleep()` call with `Delay()`

internal/latency/latency.go
```diff
@@ -0,0 +1,97 @@
...
+	return lm.locs[id-1]
+}
+
+// Enabled returns true if a latency matrix was provided.
+func (lm Matrix) Enabled() bool {
+	return lm.enabled
+}
+
+// Delay sleeps for the duration of the latency between nodes a and b.
+func (lm Matrix) Delay(a, b hotstuff.ID) {
```
AlanRostem, line 91: Suggestion: remove this method entirely and implement a separate event delaying logic in the even loop based on tickers. 

This may be expressed as a seprate issue/PR.

internal/root/dir.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"path/filepath"
+	"runtime"
+)
+
+var (
+	_, b, _, _ = runtime.Caller(0)
+
+	// Dir is the root directory of the project.
+	Dir = filepath.Join(filepath.Dir(b), "../..")
```
AlanRostem, line 12: What is this used for?

AlanRostem:
This is a much more concise approach to induced latency from various node locations. Decoupling the latency matrix logic into a struct helps create a cleaner design and using an external `.csv` file to generate the latency matrix avoids the need to hard-code the values. Some minor improvements can be made, but are not strictly necessary. 

I approve this PR for a merge, but would like others to review it and would like to be clarified further on the `internal/root/dir.go` file before we merge it to master. Also, as in the above comment, I suggest a separate PR for an improved delay induction design.

AlanRostem (approved):
Approved PR with a minor change. New issue opened: #140

PR #138, opened 2024-12-22
Tree configuration

hanish520:
create trees from position slice


trees/treeconfig.go
```diff
@@ -0,0 +1,185 @@
...
+	return 0
+}
+
+// GetHeight returns the height of the replica
+func (t *Tree) GetHeight() int {
+	return t.getHeight(t.id)
+}
+
+// GetPeers returns the peers of given ID, if any.
+func (t *Tree) GetPeers(nodeID hotstuff.ID) []hotstuff.ID {
```
meling, line 153: Let's rename it to: `PeersOf(id hotstuff.ID)`. Or perhaps `NeighborsOf`?

Is this meant to return both the `parent` and the `children`? Right now, I think it only returns the children.

And if `nodeID` is the root, shouldn't this return its children?

trees/treeconfig.go
```diff
@@ -0,0 +1,185 @@
...
+		if t.isWithInIndex(childPos) {
+			children = append(children, t.posToIDMapping[childPos])
+		} else {
+			break
+		}
+	}
+	return children
+}
+
+// getHeight returns the height of a given replica.
```
meling, line 126:
```go
// getHeight returns the height from the given replica's vantage point.
```

trees/treeconfig.go
```diff
@@ -0,0 +1,185 @@
...
+		endLimit = endLimit + int(math.Pow(float64(t.branchFactor), float64(i)))
+		if nodePos >= startLimit && nodePos <= endLimit {
+			return t.height - i
+		}
+	}
+	return 0
+}
+
+// GetHeight returns the height of the replica
+func (t *Tree) GetHeight() int {
```
meling, line 148: Do we need to export this? That is, will we use it in the code outside the `tree` package?

trees/treeconfig.go
```diff
@@ -0,0 +1,185 @@
...
+	parent, ok := t.GetParent()
+	if !ok {
+		return peers
+	}
+	parentChildren := t.GetChildrenOfNode(parent)
+	return parentChildren
+}
+
+// GetSubTreeNodes returns all the nodes of its subtree.
+func (t *Tree) GetSubTreeNodes() []hotstuff.ID {
```
meling, line 167: Let's rename it to `SubTree()`.
```go
// SubTree returns the subtree of the tree's replica.
```

trees/treeconfig.go
```diff
@@ -0,0 +1,185 @@
...
+	nodeID := t.id
+	subTreeNodes := make([]hotstuff.ID, 0)
+	children := t.GetChildrenOfNode(nodeID)
+	queue := make([]hotstuff.ID, 0)
+	queue = append(queue, children...)
+	subTreeNodes = append(subTreeNodes, children...)
+	if len(children) == 0 {
+		return subTreeNodes
```
meling, line 175: This will also be empty slice, if `len(children) == 0`

trees/treeconfig.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	for len(queue) > 0 {
```
meling, line 177: I'm a bit confused by the logic. I need to dig deeper. But perhaps it would make sense to use the std lib `slices` package for some of these things?

Maybe you could add a comment or two to explain the idea.

meling (changes requested):

trees/treeconfig.go
```diff
@@ -0,0 +1,185 @@
+package trees
```
meling, line 1: We should call it just `tree`. And we should move it to the internal package: `internal/tree`

trees/treeconfig.go
hanish520: It only returns the siblings of the replica, along with itself

trees/treeconfig.go
hanish520: This was useful to me for timer computation

trees/treeconfig.go
hanish520: no children means, it is a leaf node hence empty subtree

internal/tree/treeconfig_test.go
```diff
@@ -0,0 +1,150 @@
...
+		branchFactor      int
+		wantTreeHeight    int
+		wantChildren      []hotstuff.ID
+		wantSubTree       []hotstuff.ID
+		wantParent        hotstuff.ID
+		wantIsRoot        bool
+		wantReplicaHeight int
+		wantPeers         []hotstuff.ID
+	}{
+		{10, 1, 2, 4, []hotstuff.ID{2, 3}, []hotstuff.ID{2, 3, 4, 5, 6, 7, 8, 9, 10}, 1, true, 4, []hotstuff.ID{}},
```
meling, line 71: This table is hard to use/edit because the struct contains too many fields. Moreover, the fields should also be named on each line in the table (see examples on Line 27-33).

This test should be split up into separate tests for each function with its own anonymous `tests := struct[]{`:

```go
func TestTreeHeight  
func TestReplicaChildren  
func TestChildrenOf  
func TestParent  
func TestSubTree  
```
etc. 

By splitting up the tests, you also avoid long unlabeled table entries which are very hard to decode visually.

meling (changes requested):
I had a bunch of comments that I had forgotten to submit a review for. Sorry about that. I fixed most of them already just now, but there are two remaining issues. Could you please fix them.

internal/tree/treeconfig.go
```diff
@@ -0,0 +1,169 @@
+package tree
```
meling, line 1: Should we rename the files to just `tree.go` and `tree_test.go`?

PR #139, opened 2024-12-22
Cue-based experiment configurations

meling:
The cue-based configuration allows to specify separate machines
for clients and replicas. Further, it allows to specify how many
replicas and clients should be placed on those machines. The Go
code takes care of assigning replicas and clients to machines
based on the number of replicas and clients specified in the config.
Furthermore, you may optionally specify locations where the replicas
should be placed (and corresponding latencies will be applied by
the latency framework). Finally, you may also specify treePositions
for the replicas; the treePositions refers to the positions in the
locations list, and follow a given format (root, left, right, left-left...)
Finally, you may also specify Byzantine strategies of the different
replicas in the locations list, by providing a map from strategy
name (slow, silent) to a list of replicas that exhibit this behavior.

Please see the schema.cue for specification of constraints, and the
code for details on how to use it within the framework.


meling:
We probably need to remove the uniqueness requirement on the locations slice.

AlanRostem:
Reviewing this now. How does the tree positions work? Is this related to a simulated network topology?

AlanRostem (approved):
Cue-based config tests work, are thorough, and the changes do not conflict with the existing config setup. 

As we discussed, this needs to be adapted to the experimentation framework in a future PR.

PR #149, opened 2025-01-06
Fix #148 - Location config bug

### LINKED ISSUE #148: Location config bug
Tested the newly implemented latency matrix that was merged last week. A bug in the location config from the controller causes hosts to receive incomplete latency matrices.

I am opening this issue to fix them myself.

To elaborate, the following line updates the location list: https://github.com/relab/hotstuff/blob/503c7e78fe082141f6d54e2a623bcaebc0044b45/internal/orchestration/controller.go#L277

Since it is both being updated and added to a host in the loop, the location list is incomplete when arriving at the hosts.

AlanRostem:
Tested on bbchain-cluster with `bb1` as controller and it works like intended. Using the following configuration:

```
clients = 2
replicas = 4

hosts = [
        "bb2",
        "bb3",
        "bb4",
        "bb5",
        "bb6",
]

# specific assignments for some hosts
[[hosts-config]]
name = "bb2"
clients = 2
replicas = 0
location = "Melbourne"

[[hosts-config]]
name = "bb3"
clients = 0
replicas = 1
location = "Rome"

[[hosts-config]]
name = "bb4"
clients = 0
replicas = 1
location = "Amsterdam"

[[hosts-config]]
name = "bb5"
clients = 0
replicas = 1
location = "Barcelona"

[[hosts-config]]
name = "bb6"
clients = 0
replicas = 1
location = "Gothenburg"
```

The additional `scripts/local_config.toml` can also be used to run local tests, but the latency values will be zero.

### DIFF
--- a/internal/latency/latency.go
+++ b/internal/latency/latency.go
@@ -19,7 +19,7 @@ func Between(a, b string) time.Duration {
 
 // ValidLocation returns the location if it is valid, or an error otherwise.
 func ValidLocation(location string) (string, error) {
-	if location == "" {
+	if location == "" || location == DefaultLocation {
 		return DefaultLocation, nil
 	}
 	if !slices.Contains(allLocations, location) {
--- a/internal/latency/latency.go
+++ b/internal/latency/latency.go
@@ -67,9 +67,6 @@ func MatrixFrom(locations []string) Matrix {
 // Latency returns the latency between nodes a and b.
 // If a or b are not valid nodes, the function will panic.
 func (lm Matrix) Latency(a, b hotstuff.ID) time.Duration {
-	if a == b {
-		return 0
-	}
 	return lm.lm[a-1][b-1]
 }
 
--- a/internal/latency/latency.go
+++ b/internal/latency/latency.go
@@ -38,12 +38,6 @@ type Matrix struct {
 // MatrixFrom returns the latencies between the given locations.
 // If a location is not valid, the function will panic.
 func MatrixFrom(locations []string) Matrix {
-	if len(locations) == 0 {
-		return Matrix{
-			enabled: false,
-		}
-	}
-
 	locationIndices := make([]int, len(locations))
 	for i, loc := range locations {
 		if loc == DefaultLocation {
--- a/internal/latency/latency.go
+++ b/internal/latency/latency.go
@@ -38,6 +38,12 @@ type Matrix struct {
 // MatrixFrom returns the latencies between the given locations.
 // If a location is not valid, the function will panic.
 func MatrixFrom(locations []string) Matrix {
+	if len(locations) == 0 {
+		return Matrix{
+			enabled: false,
+		}
+	}
+
 	locationIndices := make([]int, len(locations))
 	for i, loc := range locations {
 		if loc == DefaultLocation {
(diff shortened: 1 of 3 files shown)


internal/latency/latency.go
```diff
@@ -67,6 +67,9 @@ func MatrixFrom(locations []string) Matrix {
 // Latency returns the latency between nodes a and b.
 // If a or b are not valid nodes, the function will panic.
 func (lm Matrix) Latency(a, b hotstuff.ID) time.Duration {
+	if a == b {
```
meling, line 70: Is this necessary? I think the latency matrix already has a 0 value for this entry?

internal/orchestration/controller.go
```diff
@@ -276,11 +278,11 @@ func (e *Experiment) assignReplicasAndClients() (err error) {
 			replicaOpts.ByzantineStrategy = byzantineStrategy
 			replicaLocations = append(replicaLocations, location)
 			// all replicaOpts share the same Locations slice, which is progressively updated
-			replicaOpts.Locations = replicaLocations
```
meling, line 279: Just trying to understand. Didn't this progressively updating of each `replicaOps` with the `replicaLocations` slice work?

I agree this logic is unintuitive and I don't mind this change, but wasn't sure if this was fixing a problem or just making the code logic clearer?

This stuff will probably change substantially when we start to use the code from the `cue-config` PR.

Maybe after merging this PR we should consider merging the `cue-config` PR and move forward with updating the experiment framework using the new config struct.

internal/orchestration/controller.go
```diff
@@ -276,11 +278,11 @@ func (e *Experiment) assignReplicasAndClients() (err error) {
...
 			e.hostsToReplicas[host] = append(e.hostsToReplicas[host], nextReplicaID)
 			e.replicaOpts[nextReplicaID] = replicaOpts
 			e.Logger.Infof("replica %d assigned to host %s", nextReplicaID, host)
 			nextReplicaID++
+			totalReplicaCount++
```
meling, line 285: You don't need this variable; you can just use `numReplicas`

internal/orchestration/controller.go
AlanRostem: This progressive update does not work with the new latency matrix implementation. Say you have 4 replicas, then replica 1 will only have one location in its list. Since this location list is required for creating the matrix, replica 1 won't have a large enough matrix to find the correct latency value for the others. I found this out by running the master branch version and got several indexing errors.

meling: Thanks for the clarification. Since we no longer use progressive updating of `replicaOpts.Locations` we should remove the comment about this.

internal/orchestration/controller.go
meling: My bad. I see now `totalReplicaCount` is used outside the loop where `numReplicas` is used.

AlanRostem:
Since we will change the config system anyway, I added a hacky fix to the lint error in https://github.com/relab/hotstuff/pull/149/commits/a5a39196aa115d06418d3e933d50fab9a39d57a7

Did this so the PR passes to allow merging.

meling:
This is looking good. One minor thing. Could you add a comment to this PR to explain how you test the experiment framework with the included `local_config.toml`? Or perhaps add a make target?

meling (approved):
LGTM!

PR #151, opened 2025-01-07
Update to 1.24rc1. Updated benchmark tool usage.

### LINKED ISSUE #147: Migrate to Go 1.24rc1
A release candidate for Go 1.24 is available. For now, the following enhancements can be utilized:

- b.Loop() instead of b.N in benchmarks
- synctest package to speed up tests using functions like time.Sleep()

AlanRostem:
Fixes #147

Runs locally without issues, test pass and benchmarks work as intended.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/relab/hotstuff
 
-go 1.23.4
+go 1.24rc1
 
 require (
 	github.com/felixge/fgprof v0.9.5
--- a/modules/options_test.go
+++ b/modules/options_test.go
@@ -20,7 +20,7 @@ func BenchmarkOptionsGet(b *testing.B) {
 	opts := modules.Options{}
 	var v any
 	opts.Set(TestOption, 1)
-	for i := 0; i < b.N; i++ {
+	for b.Loop() {
 		v = opts.Get(TestOption).(int)
 	}
 	_ = v
@@ -30,7 +30,7 @@ func BenchmarkOptionsField(b *testing.B) {
 	opts := modules.Options{}
 	var v any
 	opts.SetSharedRandomSeed(1)
-	for i := 0; i < b.N; i++ {
+	for b.Loop() {
 		v = opts.SharedRandomSeed()
 	}
 	_ = v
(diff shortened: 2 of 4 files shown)


modules/options_test.go
```diff
@@ -20,7 +20,7 @@ func BenchmarkOptionsGet(b *testing.B) {
 	opts := modules.Options{}
 	var v any
 	opts.Set(TestOption, 1)
-	for i := 0; i < b.N; i++ {
+	for b.Loop() {
 		v = opts.Get(TestOption).(int)
```
meling, line 24: Here, you can remove the `v` variable altogether since the reason for `b.Loop()` is to tell the compiler not to optimize away results (and code inside the benchmark loop). The same goes for the next benchmark below.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/relab/hotstuff
 
-go 1.23.4
+go 1.24rc1
```
meling, line 3: I forgot to mention that when updating to newer versions of Go, we must also update GitHub actions workflow files that refer to the version to test. I'm not sure the different actions work with release candidate versions, but we should try. But I am sure that `golangci-lint` will not work with newer versions of Go unless we compile our own `golangci-lint` binary, and that's a pain not worth doing in this PR.

meling (changes requested):
Needs tweaking in github workflow files, but don't waste time on `golangci-lint`.

modules/options_test.go
AlanRostem: Would something like this suffice?

```
func BenchmarkOptionsGet(b *testing.B) {
	opts := modules.Options{}
	opts.Set(TestOption, 1)
	for b.Loop() {
		opts.Get(TestOption)
	}
}
```

meling: I believe so, but please double-check that the benchmark gives similar results with both alternatives.

AlanRostem: Checked and both alternatives are similar

meling:
Remember also to update the `Makefile`.

go.mod
AlanRostem: Tried it and it seems to not work. Maybe we need to wait for the full release?

meling: Yeah, let's not waste time on this.

PR #155, opened 2025-01-09
Cue config integration

### LINKED ISSUE #153: Make experiment framework compatible with Cue-based config implementation
The Cue configs are not being used by the experiment framework. Integrating the Cue config implementation simplifies a lot of code on top of built-in config validation.

This issue is related to #150 in that the experiment framework can be simplified this way.

AlanRostem:
Fixes #153 

Works both locally and distributed. As a result of integrating Cue, `internal/orchestration/controller.go` and `internal/cli/run.go` have been simplified. 

The byzantine strategies have not been tested yet.

Feel free to review the code and write suggestions.


internal/config/config.go
```diff
@@ -103,3 +122,13 @@ func (c *Config) AssignClients() map[string][]hotstuff.ID {
 	}
 	return hostsToClients
 }
+
+// Returns true if both the replica and client hosts slices
```
meling, line 126: It is common (unless recommendations have changed) to start doc comments with the name of the func.
```go
// IsLocal returns true ...
```

internal/config/config.go
```diff
@@ -103,3 +122,13 @@ func (c *Config) AssignClients() map[string][]hotstuff.ID {
...
+// contain one instance of "localhost".
+func (c *HostConfig) IsLocal() bool {
+	if len(c.ClientHosts) > 1 || len(c.ReplicaHosts) > 1 {
+		return false
+	}
+
```
meling, line 132: Nit: I prefer to avoid empty lines in short methods, especially if there is a single `}` (almost empty line) above it.

internal/orchestration/controller.go
```diff
@@ -7,50 +7,74 @@ import (
...
+		duration:    duration,
+		output:      outputDir,
+	}
+}
+
+func (e *Experiment) SetWorkers(hosts map[string]RemoteWorker) {
+	e.workers = hosts
+}
+
+func (e *Experiment) SetHostConfig(cfg *config.HostConfig) error {
```
meling, line 68: Maybe this could be done in the `NewExperiment` func above?

internal/orchestration/controller.go
```diff
@@ -62,112 +86,101 @@ func (e *Experiment) Run() (err error) {
...
+	err = e.stopReplicas(replicaMap)
 	if err != nil {
 		return fmt.Errorf("failed to stop replicas: %w", err)
 	}
 
 	return nil
 }
 
-func (e *Experiment) createReplicas() (cfg *orchestrationpb.ReplicaConfiguration, err error) {
+func (e *Experiment) createReplicas(m config.ReplicaMap) (cfg *orchestrationpb.ReplicaConfiguration, err error) {
```
meling, line 139: Prefer more descriptive name for the `ReplicaMap` variable: `s/m/replicaMap`

internal/orchestration/controller.go
```diff
@@ -62,112 +86,101 @@ func (e *Experiment) Run() (err error) {
...
-
+	for host, opts := range m {
+		worker := e.workers[host]
 		req := &orchestrationpb.CreateReplicaRequest{Replicas: make(map[uint32]*orchestrationpb.ReplicaOpts)}
-		for _, id := range e.hostsToReplicas[host] {
-			opts := e.replicaOpts[id]
-			opts.CertificateAuthority = keygen.CertToPEM(e.ca)
 
-			// the generated certificate should be valid for the hostname and its ip addresses.
+		for _, opt := range opts {
```
meling, line 151: The loop content is a great target for refactoring to a separate function.

internal/orchestration/controller.go
```diff
@@ -176,149 +189,15 @@ func (e *Experiment) createReplicas() (cfg *orchestrationpb.ReplicaConfiguration
...
 type assignmentsFileContents struct {
 	// the host associated with each replica.
 	HostsToReplicas map[string][]hotstuff.ID
 	// the host associated with each client.
 	HostsToClients map[string][]hotstuff.ID
 }
 
-func (e *Experiment) writeAssignmentsFile() (err error) {
-	f, err := os.OpenFile(filepath.Join(e.Output, "hosts.json"), os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0o644)
+func (e *Experiment) writeAssignmentsFile(m config.ReplicaMap, clientIDs config.ClienIdMap) (err error) {
```
meling, line 199: Same here; `m` is not the best name.

internal/orchestration/controller.go
```diff
@@ -329,34 +208,43 @@ func (e *Experiment) writeAssignmentsFile() (err error) {
...
+		for _, opt := range opts {
+			replicaIDs[host] = append(replicaIDs[host], opt.HotstuffID())
+		}
+	}
+
 	return enc.Encode(assignmentsFileContents{
-		HostsToReplicas: e.hostsToReplicas,
-		HostsToClients:  e.hostsToClients,
+		HostsToReplicas: replicaIDs,
+		HostsToClients:  clientIDs,
```
meling, line 222: Reminder: Consider what needs to be written to the `hosts.json` file, and should this be a more comprehensive format with more experiment details.

internal/config/config.go
```diff
@@ -89,9 +106,11 @@ func (c *Config) lookupByzStrategy(replicaID hotstuff.ID) string {
 	return ""
 }
 
+type ClienIdMap map[string][]hotstuff.ID
```
meling, line 109: Should we name it `ClientMap`? Should it too work with `orchestrationpb.ClientOpts` similar to `AssignReplicas`? Not sure...

internal/orchestration/controller.go
```diff
@@ -430,10 +318,18 @@ func (e *Experiment) quit() error {
...
 	var ids []uint32
-	for _, id := range m[host] {
-		ids = append(ids, uint32(id))
+	for _, opts := range m[host] {
+		ids = append(ids, uint32(opts.ID))
 	}
 	return ids
 }
+
+func clientIDsToU32(host string, clientIDs config.ClienIdMap) []uint32 {
```
meling, line 329: Could we avoid this helper by using `ClientOpts`?

.vscode/launch.json
```diff
@@ -12,7 +12,8 @@
 			"program": "${workspaceRoot}/cmd/hotstuff/main.go",
 			"cwd": "${workspaceRoot}",
 			"args": [
-				"run"
+				"run",
+				"--config=scripts/local_config.cue"
```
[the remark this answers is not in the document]
AlanRostem: Do you mean a doc on how to use it in general? Sure.

This line can actually be removed since I just used it to get the debug breaks in vscode. A default local config is already being generated when one is not specified.

internal/orchestration/controller.go
AlanRostem: Yes it should be in the constructor, same with `SetWorkers`. 

Though I'd like your opinion on constructors with a large amount of parameters. This function is quite "long"

AlanRostem:
@meling Made the following changes in the commit above, based on your comments:

- Rename `m` to be more descriptive
- Refactor the contents of the `for _, opt := range opts` loop to a function `putReplicaOpt`
- Reduced the number of params in `NewExperiment` and `Experiment` no longer holds ǹumReplicas` and `numClients`. We are using the `e.hostCfg` for that.
- Removed the setters in `Experiment` and supply the values in `NewExperiment` instead

Still not sure how to avoid `clientIdsToU32`, but I am thinking about it

internal/orchestration/controller.go
```diff
@@ -62,112 +77,109 @@ func (e *Experiment) Run() (err error) {
...
-			}
-
-			// add the internal address as well
-			if internalAddr != "" {
-				validFor = append(validFor, internalAddr)
-			}
 
-			keyChain, err := keygen.GenerateKeyChain(id, validFor, e.Crypto, e.ca, e.caKey)
+		for _, opt := range opts {
+			err = e.putReplicaOpt(req.Replicas, host, opt)
```
meling, line 169: Here is a two-step suggestion:
1. Rename and remove the first argument and update the call site:
```go
func (e *Experiment) setReplicaCertificates(host string, opt *orchestrationpb.ReplicaOpts) error {
...
	opt.CertificateKey = keyChain.CertificateKey
// Don't update map here.
	return nil
}
```
```go
	for host, opts := range replicaMap {
		worker := e.workers[host]
		replicaMap := make(map[uint32]*orchestrationpb.ReplicaOpts)
		for _, opt := range opts {
			if err = e.setReplicaCertificates(host, opt); err != nil {
				return nil, err
			}
			replicaMap[opt.ID] = opt
		}

		req := &orchestrationpb.CreateReplicaRequest{Replicas: replicaMap}
		wcfg, err := worker.CreateReplica(req)
...
```
2. Move the `setReplicaCertificates` to be a method on the `orchestrationpb.ReplicaOpts` type instead of `Experiment`. I have already added a file for this `internal/proto/orchestrationpb/replica_opts.go`.
```go
func (x *ReplicaOpts) SetReplicaCertificates(host string, ca *x509.Certificate, caKey *ecdsa.PrivateKey) error {
	x.CertificateAuthority = keygen.CertToPEM(ca)
	validFor := []string{"localhost", "127.0.0.1", "127.0.1.1", host}
	ips, err := net.LookupIP(host)
	if err == nil {
		for _, ip := range ips {
			// TODO: Not using internal addr anymore, but check if this is needed.
			if ipStr := ip.String(); ipStr != host /*&& ipStr != internalAddr*/ {
				validFor = append(validFor, ipStr)
			}
		}
	}

	keyChain, err := keygen.GenerateKeyChain(hotstuff.ID(x.ID), validFor, x.Crypto, ca, caKey)
	if err != nil {
		return fmt.Errorf("failed to generate keychain: %w", err)
	}
	x.PrivateKey = keyChain.PrivateKey
	x.PublicKey = keyChain.PublicKey
	x.Certificate = keyChain.Certificate
	x.CertificateKey = keyChain.CertificateKey
	return nil
}
```

internal/orchestration/controller.go
```diff
@@ -430,10 +317,18 @@ func (e *Experiment) quit() error {
 	return nil
 }
 
-func getIDs(host string, m map[string][]hotstuff.ID) []uint32 {
+func replicaIDsToU32(host string, replicaMap config.ReplicaMap) []uint32 {
```
meling, line 320: This can be replaced with this one (in `config/config.go`):
```go
func (r ReplicaMap) ReplicaIDs(host string) []uint32 {
	ids := make([]uint32, 0, len(r[host]))
	for _, opts := range r[host] {
		ids = append(ids, opts.ID)
	}
	return ids
}
```
And similar for the `ClientMap`.

meling (changes requested):
Added a few more suggestions...

internal/config/config.go
```diff
@@ -34,6 +34,23 @@ type Config struct {
 	ByzantineStrategy map[string][]uint32
 }
 
+// NewLocal creates a config for a localhost case.
+// TODO: Add support for locations through cli.
```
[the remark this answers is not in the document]
AlanRostem: Yes, I think it would be a nice option.

AlanRostem:
@meling The commit above adds your suggestions.

Also, shouldn't `controller.go` be renamed to `experiment.go` considering the all code is related to the `Experiment` struct now?

We could also rename `Experiment` to `Controller`

internal/cli/run.go
```diff
@@ -98,43 +99,23 @@ func runController() {
...
+		cfg, err = config.Load(cfgPath)
+		checkf("config error: %v", err)
+	} else {
+		cfg = config.NewLocal(numReplicas, numClients)
+	}
 
 	worker := viper.GetBool("worker")
-	hosts := viper.GetStringSlice("hosts")
+
+	hosts := cfg.AllHosts()
```
meling, line 117: This returns an empty slice if the cfg (HostConfig) has localhost. Is that intended?

hotstuff.go
```diff
@@ -12,3 +12,16 @@ func (id ID) ToBytes() []byte {
 	binary.LittleEndian.PutUint32(idBytes[:], uint32(id))
 	return idBytes[:]
 }
+
+// IDSlice is a slice containing IDs.
+type IDSlice []ID
+
+// Uint32Slice converts the IDSlice to a []uint32.
+// Useful for passing into protocol buffer types.
+func (ids IDSlice) Uint32Slice() []uint32 {
```
meling, line 21: Not sure we need this. See other comments.

internal/config/config.go
```diff
@@ -7,8 +7,27 @@ import (
 	"github.com/relab/hotstuff/internal/proto/orchestrationpb"
 )
 
-// Config holds the configuration for an experiment.
-type Config struct {
+// ReplicaMap is a map from host to a list of replica options.
+type ReplicaMap map[string][]*orchestrationpb.ReplicaOpts
+
+// ReplicaIDs converts the an entry map from []hotstuff.ID to []uint32.
+func (r ReplicaMap) ReplicaIDs(host string) hotstuff.IDSlice {
```
meling, line 14: This method should probably return a `[]uint32` directly, eliminating the need to call the conversion function later.

internal/config/config.go
```diff
@@ -7,8 +7,27 @@ import (
...
+	for _, opts := range r[host] {
+		ids = append(ids, hotstuff.ID(opts.ID))
+	}
+	return ids
+}
+
+type ClientMap map[string][]hotstuff.ID
+
+// ClientIDs converts the an entry map from []hotstuff.ID to []uint32.
+func (c ClientMap) ClientIDs(host string) hotstuff.IDSlice {
```
meling, line 25: This method should probably return a `[]uint32` directly, eliminating the need to call the conversion function later.

internal/config/config.go
```diff
@@ -103,3 +144,22 @@ func (c *Config) AssignClients() map[string][]hotstuff.ID {
...
+	}
+	return c.ReplicaHosts[0] == "localhost" && c.ClientHosts[0] == "localhost" ||
+		c.ReplicaHosts[0] == "127.0.0.1" && c.ClientHosts[0] == "127.0.0.1"
+}
+
+// AllHosts returns the list of all hostnames, including replicas and clients.
+// If the configuration is set to run locally, the function returns an empty list.
+func (c *HostConfig) AllHosts() []string {
+	if c.isLocal() {
+		return []string{}
```
meling, line 162: I think this might not be what we want to return for localhost, or did I miss something?

meling (changes requested):
Submitting now before I forget that I have comments here... I may have missed some conceptual idea with the IDSlice (noticed only late that it is also changed in tree package.

internal/config/config.go
AlanRostem: I did this because for some reason having a slice with one "localhost" in it causes an error that starts with this line:

https://github.com/relab/hotstuff/blob/a811b3b6c3aee4da759f6a3a1ca27542c123de6d/internal/cli/run.go#L121

the outputs the following when running locally: 

```
% ./hotstuff run
run.go:219: failed to connect to remote hosts: dial tcp [::1]:22: connect: connection refused
```

Making the slice empty is a workaround to this error. Not sure of other ways to solve this

AlanRostem:
Discovered the problem with nil reference in `crypto.VerifyQuorumCert()`: For some reason when running exactly two byzantine silent nodes and one that forks, the `signature` field in the `hotstuff.QuorumCert` struct. Here is a config that reproduces the problem: 

```
package config

config: {
        replicaHosts: [
                "bb2",
                "bb3",
                "bb4",
                "bb5",
                "bb6",
                "bb7",
        ]
        clientHosts: [
                "bb8",
                "bb9",
        ]

        replicas: 10
        clients:  2


        byzantineStrategy: {
                silence: [2, 4]
                fork: [5]
        }
}

```

...and here is the exact line where the error occurs: 

https://github.com/relab/hotstuff/blob/4102739d618d6f7827e8ded51ffe5274a6c9c460/crypto/crypto.go#L107

...with modifications below to get the exact nil reference:

```
// VerifyQuorumCert verifies a quorum certificate.
func (c crypto) VerifyQuorumCert(qc hotstuff.QuorumCert) bool {
	// genesis QC is always valid.
	if qc.BlockHash() == hotstuff.GetGenesis().Hash() {
		return true
	}
	signature := qc.Signature() // <-- returns nil
	participants := signature.Participants()
	if participants.Len() < c.configuration.QuorumSize() {
		return false
	}
	block, ok := c.blockChain.Get(qc.BlockHash())
	if !ok {
		return false
	}
	return c.Verify(qc.Signature(), block.ToBytes())
}
``` 

Still investigating the root cause, since this oddly only happens using the above config. Would it be a good fix to check if the signature is nil and return false in the above method?

AlanRostem:
Upon rigorously debugging the nil reference issue, I discovered that this bug is also on the master branch. I wasn't able to get the `.toml` config to work like the `.cue` config shown in the above comment, so I had to hard-code the byzantine values to the exact replica IDs above. 

Adding the below changes to this line: https://github.com/relab/hotstuff/blob/4102739d618d6f7827e8ded51ffe5274a6c9c460/internal/orchestration/controller.go#L277

```go
	if replicaOpts.ID == 2 || replicaOpts.ID == 4 {
		byzantineStrategy = "silence"
	} else if replicaOpts.ID == 5 {
		byzantineStrategy = "fork"
	}
```

while running this command on the cluster: `./hotstuff run --replicas=6 --clients=2 --hosts=bb2,bb3,bb4,bb5,bb6,bb7,bb8,bb9`

Reproduces the error, meaning that this bug is not related to Cue and I believe we can merge this branch to master.

Still, this is a very strange bug that seems to only trigger with this **specific** configuration of byzantine strategies in the replicas. I checked it and it seems like when one of these are the leader, they send a proposal with a nil signature. Is this intended?

meling:
> Still, this is a very strange bug that seems to only trigger with this **specific** configuration of byzantine strategies in the replicas. I checked it and it seems like when one of these are the leader, they send a proposal with a nil signature. Is this intended?

Maybe the leader is Byzantine ;-)

Joking aside; I think I've been debugging this same problem as well... at least it looks familiar. I will to take a look when I get a chance...

AlanRostem:
I am working on merging the Cue config integration branch and I was wondering how Kauri should be tested? I am not seeing the `modules.Kauri` being added to the builder in `worker.go`.

kauri/kauri_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"testing"
+
+	"github.com/relab/hotstuff"
+	"github.com/relab/hotstuff/crypto"
+	"github.com/relab/hotstuff/crypto/ecdsa"
+)
+
+func TestSubTree(t *testing.T) {
+	subTreeTestData := []struct {
+		test_a   []hotstuff.ID
```
meling, line 13: This stutters when used `test.test_a`. Change to just `a` and `b`...

I prefer the shorter `want` instead of `expected`.

kauri/kauri_test.go
```diff
@@ -0,0 +1,64 @@
...
+			t.Errorf("SubTree(%v, %v) = %v; want %v", test.test_a, test.test_b, !test.expected, test.expected)
+		}
+	}
+}
+
+func TestCanMerge(t *testing.T) {
+	mergeTestData := []struct {
+		test_a   []hotstuff.ID
+		test_b   []hotstuff.ID
+		expected bool
```
meling, line 34: Same for `a` and `b` as comment above. Replace with `wantErr` to more clearly signify that the bool represents that we want an error or not. See other examples; we should try to follow the same pattern.

meling (changes requested):

PR #156, opened 2025-01-09
Omit Handel Aggregation protocol

### LINKED ISSUE #142: Problems with "TestOrchestration/*+Handel" tests
All TestOrchestration tests with a *+Handel* postfix pass but the number of executed commands become zero. The output of all such tests are the following:

`2025-01-03T13:03:00.201Z	INFO	cli2	client/client.go:170	Done sending commands (executed: 0, failed: 0, timeouts: 2268)`

Is this intended behavior with the Handel scheme? If not, this should be investigated and resolved. 

Considering that the last changes made to Handel was 3 years ago, it may be desirable to omit this protocol altogether.

AlanRostem:
Fixes #142

As we discussed, this aggregation protocol is not needed in the framework and can be removed.


hanish520 (approved):
looks good to me

meling (approved):

PR #158, opened 2025-01-12
Kauri-impl

hanish520:
Porting Kauri implementation from previous implementation


kauri/kauri.go
```diff
@@ -0,0 +1,301 @@
...
+	ids := make([]hotstuff.ID, 0, totalNodes)
+	for id := range k.configuration.Replicas() {
+		ids = append(ids, id)
+	}
+	sort.Slice(ids, func(i, j int) bool { return ids[i] < ids[j] })
+	seed := k.opts.SharedRandomSeed() + int64(binary.LittleEndian.Uint64(hash[:]))
+	//Shuffle the list of IDs using the shared random seed + the first 8 bytes of the hash.
+	rnd := rand.New(rand.NewSource(seed))
+	rnd.Shuffle(len(ids), reflect.Swapper(ids))
+	lIndex := 0
```
meling, line 287: Finding the leader index should use std lib func:
```go
lIndex := slices.Index(ids, leaderID)
if lIndex == -1 {
	panic("leader is not in the list of IDs")
}
```
The question is what to do if someone passes a leader that's not in the `ids` set.

kauri/kauri.go
```diff
@@ -0,0 +1,301 @@
...
+	for index, id := range ids {
+		if id == leaderID {
+			lIndex = index
+		}
+	}
+	currentRoot := ids[0]
+	ids[0] = ids[lIndex]
+	ids[lIndex] = currentRoot
+	posMapping := make(map[hotstuff.ID]int)
```
meling, line 296: Why not just return the `ids` slice? Why do you want it as a map?

Since the hotstuff.IDs are 1-indexed, it would be possible to use a slice where the first entry is ignored, then you could do:
```go
idx := posMapping[myID]
```
To get the index of the `hotstuff.ID myID` in the same way as with a map.
Code would be like this:
```go
pos := make([]int, len(ids)+1)
for index, id := range ids {
	pos[id] = index
}
```

meling (changes requested):

kauri/kauri.go
```diff
@@ -0,0 +1,268 @@
...
+func (i serviceImpl) SendContribution(_ gorums.ServerCtx, request *kauripb.Contribution) {
+	i.k.eventLoop.AddEvent(ContributionRecvEvent{Contribution: request})
+}
+
+// ContributionRecvEvent is raised when a contribution is received.
+type ContributionRecvEvent struct {
+	Contribution *kauripb.Contribution
+}
+
+func (k *Kauri) canMergeContributions(a, b hotstuff.QuorumSignature) bool {
```
meling, line 189: I believe the following logic is a bit clearer:
```go
func (k *Kauri) canMergeContributions(a, b hotstuff.QuorumSignature) bool {
	if a == nil || b == nil {
		k.logger.Info("kauri: cannot merge nil contributions")
		return false
	}
	a.Participants().RangeWhile(func(i hotstuff.ID) bool {
		// cannot merge a and b if both contain a contribution from the same ID.
		return !b.Participants().Contains(i)
	})
	return true
}
```
But I would be more confident in my statement if we had a test for this one. Could you please add a test?

kauri/kauri.go
```diff
@@ -0,0 +1,268 @@
...
+				canMerge = false
+			}
+			return canMerge
+		})
+		return canMerge
+	})
+	return canMerge
+}
+
+func (k *Kauri) verifyContribution(signature hotstuff.QuorumSignature, hash hotstuff.Hash) bool {
```
meling, line 208:
```go
func (k *Kauri) verifyContribution(signature hotstuff.QuorumSignature, hash hotstuff.Hash) bool {
	block, ok := k.blockChain.Get(hash)
	if !ok {
		k.logger.Info("kauri: failed to fetch the block", hash)
		return false
	}
	return k.crypto.Verify(signature, block.ToBytes())
}
```

meling (changes requested):
I'm adding my comments, specifically to the Kauri code... I'll follow up with a commit and push to fix the issues myself to save time, but I'm leaving the comments for the record.

Some of them are still valid questions from my side. I especially feel we should add a few tests for some of the logic. It should be easy to test the `canMergeContributions` function (since it no longer needs the Kauri struct.)

AlanRostem:
I am working on merging this branch https://github.com/relab/hotstuff/tree/153-cue-integration

...into this Kauri branch and I was wondering how Kauri should be run? I am not seeing the `modules.Kauri` being added to the builder in `worker.go`.

PR #163, opened 2025-01-14
EventQueue fixes

### LINKED ISSUE #154: Chore: eventloop/queue implementation has some weirdness
We do this:
```go
if len(q.entries) == 0 {
	panic("cannot push to a queue with capacity 0")
}
```
in the `push` method? Moreover, this is not how to check the capacity of a slice.

If we want to check the `capacity` of the slice, we should do it when constructing the queue, using the input `capacity` argument, not after it has been created. This would be how you check the capacity of a slice, if needed:
```go
if cap(q.entries) == 0 {
	panic("cannot push to a queue with capacity 0")
}
```
If the idea was to check the `length` of the slice, then why can't you push to an empty queue with (presumably some) capacity?

AlanRostem:
Fixes #154 

This PR improves the documentation of the event queue implementation, fixes an invalid check, and adds a warning log when the event queue is full.

All tests pass and things run as intended. Though it would be nice open an issue for stress testing the event queue further.

### DIFF
--- a/eventloop/eventloop.go
+++ b/eventloop/eventloop.go
@@ -12,6 +12,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/relab/hotstuff/logging"
+	"github.com/relab/hotstuff/modules"
 	"github.com/relab/hotstuff/util/gpool"
 )
 
@@ -51,6 +53,8 @@ type handler struct {
 
 // EventLoop accepts events of any type and executes registered event handlers.
 type EventLoop struct {
+	logger logging.Logger
+
 	eventQ queue
 
 	mut sync.Mutex // protects the following:
@@ -75,6 +79,10 @@ func New(bufferSize uint) *EventLoop {
 		tickers:       make(map[int]*ticker),
 	}
 	return el
+}
+
+func (el *EventLoop) InitModule(mods *modules.Core) {
+	mods.Get(&el.logger)
 }
 
 // RegisterObserver registers a handler with priority.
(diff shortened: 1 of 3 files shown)


eventloop/queue.go
```diff
@@ -21,13 +25,13 @@ func newQueue(capacity uint) queue {
 	}
 }
 
-func (q *queue) push(entry any) {
+// push adds an entry to the buffer in a FIFO fashion. If the queue is full, the first
+// entry is dropped to make space for the newest entry and returns true.
+func (q *queue) push(entry any) (droppedOne bool) {
```
meling, line 30: What if we instead return the dropped event; then we can log the event that was dropped. See implementation below.
```go
func (q *queue) push(entry any) (droppedEvent any) {
...
		droppedEvent = q.entries[q.head] // Line 46 with current: dropped = true
...
	return droppedEvent
```

eventloop/eventloop.go
```diff
@@ -129,7 +137,10 @@ func (el *EventLoop) AddEvent(event any) {
 	if event != nil {
 		// run handlers with runInAddEvent option
 		el.processEvent(event, true)
-		el.eventQ.push(event)
+		dropped := el.eventQ.push(event)
```
meling, line 140: See details below...
```go
droppedEvent := el.eventQ.push(event)
if droppedEvent != nil {
	el.logger.Warnf("event queue is full, dropped event: %v", droppedEvent)
}
```

meling (approved):

PR #167, opened 2025-01-16
Following testing convention in TestOrchestration

### LINKED ISSUE #159: Chore: Make TestOrchestration subtests follow standard testing style
**This should wait until ongoing PRs that touch this test have been merged...**

Instead of:
```go
func TestOrchestration(t *testing.T) {
	run := func(t *testing.T, consensusImpl string, crypto string, mods []string, byzantine string) {
```
We should write:
```go
func TestOrchestration(t *testing.T) {
	tests := []struct {
		consensus string
		crypto    string
		byzantine string
		mods      []string
	}{
		{consensus: "chainedhotstuff", crypto: "ecdsa", byzantine: "", mods: nil},
		{consensus: "chainedhotstuff", crypto: "eddsa", byzantine: "", mods: nil},
		{consensus: "chainedhotstuff", crypto: "bls12", byzantine: "", mods: nil},
		{consensus: "fasthotstuff", crypto: "ecdsa", byzantine: "", mods: nil},
		{consensus: "fasthotstuff", crypto: "eddsa", byzantine: "", mods: nil},
		{consensus: "fasthotstuff", crypto: "bls12", byzantine: "", mods: nil},
		{consensus: "simplehotstuff", crypto: "ecdsa", byzantine: "", mods: nil},
		{consensus: "simplehotstuff", crypto: "eddsa", byzantine: "", mods: nil},
		{consensus: "simplehotstuff", crypto: "bls12", byzantine: "", mods: nil},
		{consensus: "chainedhotstuff", crypto: "ecdsa", byzantine: "fork:1", mods: nil},
		{consensus: "chainedhotstuff", crypto: "ecdsa", byzantine: "silence:1", mods: nil},
		{consensus: "chainedhotstuff", crypto: "ecdsa", byzantine: "", mods: []string{"handel"}},
		{consensus: "chainedhotstuff", crypto: "bls12", byzantine: "", mods: []string{"handel"}},
	}
	for _, tt := range tests {
		t.Run(test.Name([]string{"consensus", "crypto", "byzantine", "mods"}, tt.consensus, tt.crypto, tt.byzantine), func(t *testing.T) {
			// content of run function
		})
	}
```
Here is the implementation of `test.Name` (which could be placed in `internal/test` package):
```go

// Name returns the test name based on the provided fields and values.
func Name(fields []string, values ...any) string {
	if len(fields) != len(values) {
		panic("fields and values must have the same length")
	}
	b := strings.Builder{}
	for i, f := range fields {
		v := values[i]
		switch x := v.(type) {
		case []string:
			if x == nil {
				b.WriteString(fmt.Sprintf("/%s=<nil>", f))
			} else {
				b.WriteString(fmt.Sprintf("/%s=%v", f, x))
			}
		case string:
			if x != "" {
				b.WriteString(fmt.Sprintf("/%s=%s", f, v))
			}
		case uint64:
			if x != 0 {
				b.WriteString(fmt.Sprintf("/%s=%d", f, v))
			}
		case int:
			if x != 0 {
				b.WriteString(fmt.Sprintf("/%s=%d", f, v))
			}
		case bool:
			if x {
				b.WriteString(fmt.Sprintf("/%s", f))
			}
		default:
			b.WriteString(fmt.Sprintf("/%s=%v", f, v))
		}
	}
	return b.String()
}
```

AlanRostem:
Fixes #159

### DIFF
--- a/internal/orchestration/controller.go
+++ b/internal/orchestration/controller.go
@@ -153,7 +153,7 @@ func (e *Experiment) createReplicas(replicaMap config.ReplicaMap) (cfg *orchestr
 			if err != nil {
 				return nil, err
 			}
-			opt.SetTreeOptions(e.hostCfg.BranchFactor, e.hostCfg.TreePositions, e.hostCfg.TreeDelta)
+			// opt.SetTreeOptions(e.hostCfg.BranchFactor, e.hostCfg.TreePositions, e.hostCfg.TreeDelta)
 			req.Replicas[opt.ID] = opt
 			e.logger.Infof("replica %d assigned to host %s", opt.ID, host)
 		}
--- a/internal/orchestration/orchestration_test.go
+++ b/internal/orchestration/orchestration_test.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"runtime"
+	"slices"
 	"strings"
 	"sync"
 	"testing"
@@ -20,6 +21,7 @@ import (
 	"github.com/relab/hotstuff/internal/orchestration"
 	"github.com/relab/hotstuff/internal/proto/orchestrationpb"
 	"github.com/relab/hotstuff/internal/protostream"
+	"github.com/relab/hotstuff/internal/test"
 	"github.com/relab/hotstuff/internal/tree"
 	"github.com/relab/hotstuff/logging"
 	"github.com/relab/hotstuff/metrics"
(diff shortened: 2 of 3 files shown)


internal/orchestration/controller.go
```diff
@@ -153,7 +153,7 @@ func (e *Experiment) createReplicas(replicaMap config.ReplicaMap) (cfg *orchestr
 			if err != nil {
 				return nil, err
 			}
-			opt.SetTreeOptions(e.hostCfg.BranchFactor, e.hostCfg.TreePositions, e.hostCfg.TreeDelta)
+			// opt.SetTreeOptions(e.hostCfg.BranchFactor, e.hostCfg.TreePositions, e.hostCfg.TreeDelta)
```
meling, line 156: Keep or delete?

internal/orchestration/orchestration_test.go
```diff
@@ -77,85 +79,83 @@ func makeClientOpts() *orchestrationpb.ClientOpts {
...
+		{consensus: "chainedhotstuff", crypto: "bls12", byzantine: "", mods: []string{"kauri"}},
+		{consensus: "chainedhotstuff", crypto: "ecdsa", byzantine: "", mods: []string{"kauri"}, randomTree: true},
+		{consensus: "chainedhotstuff", crypto: "bls12", byzantine: "", mods: []string{"kauri"}, randomTree: true},
+	}
+
+	for _, tt := range tests {
+		t.Run(test.Name([]string{"consensus", "crypto", "byzantine", "mods"}, tt.consensus, tt.crypto, tt.byzantine, tt.mods), func(t *testing.T) {
+			var replicaOpts *orchestrationpb.ReplicaOpts
+			if slices.Contains(tt.mods, "kauri") {
+				replicaOpts = makeTreeReplicaOpts(tt.consensus, tt.crypto, tt.mods, 7, 2, tt.randomTree)
```
meling, line 152: We should make 7, 2 be part of the `tests` struct, and leave it unspecified (0) for the non-kauri tests.

internal/orchestration/orchestration_test.go
```diff
@@ -77,85 +79,83 @@ func makeClientOpts() *orchestrationpb.ClientOpts {
...
+
+func TestOrchestration(t *testing.T) {
+	tests := []struct {
+		consensus  string
+		crypto     string
+		byzantine  string
+		mods       []string
+		randomTree bool
+	}{
+		{consensus: "chainedhotstuff", crypto: "ecdsa", byzantine: "", mods: nil},
```
meling, line 131: we should remove `mods: nil` from lines that don't need it.

meling: the same applies to byzantine I think; it will be the zero value anyway.

internal/orchestration/orchestration_test.go
```diff
@@ -77,85 +79,83 @@ func makeClientOpts() *orchestrationpb.ClientOpts {
...
-			t.Fatal(err)
-		}
+func run(t *testing.T, replicaOpts *orchestrationpb.ReplicaOpts) {
+	t.Helper()
+
+	controllerStream, workerStream := net.Pipe()
+	workerProxy := orchestration.NewRemoteWorker(protostream.NewWriter(controllerStream), protostream.NewReader(controllerStream))
+	worker := orchestration.NewWorker(protostream.NewWriter(workerStream), protostream.NewReader(workerStream), metrics.NopLogger(), nil, 0)
+
+	cfg := config.NewLocal(7, 2)
```
meling, line 89: hardcoded?

meling: should these be moved into the `tests` struct?

internal/test/name.go
```diff
@@ -0,0 +1,47 @@
...
+
+// Name returns the test name based on the provided fields and values.
+func Name(fields []string, values ...any) string {
+	if len(fields) != len(values) {
+		panic("fields and values must have the same length")
+	}
+	b := strings.Builder{}
+	sep := ""
+	for i, f := range fields {
+		v := values[i]
```
meling, line 16: We should probably discuss whether or not to print the zero value or drop it. Maybe it would make sense to have a trick to encode "print even if zero-value".

meling (changes requested):
Looks very nice!! A few minor nits and one open question...

internal/orchestration/controller.go
AlanRostem: Keep! I forgot to uncomment this. On that note, we still might want to remove this line in a different branch since the tree values don't need to be set here when they can come from `replicaOpts` in `NewExperiment` instead

AlanRostem: Opened #168 to address this problem and some other stuff

meling (commented):
This one also appears to conflict heavily with #170 in orchestration_test.go. I suggest to salvage what's relevant in this one and move it to #170.

Issue #168, opened 2025-01-16
Simplify the configuration handling code


AlanRostem:
The current way the config is handled is very complex for handling both cli/viper and Cue, and difficult to maintain compatibility between them.

I suggest putting all possible config parameters in the `HostConfig` from https://github.com/relab/hotstuff/blob/master/internal/config/config.go

...and implement two constructor variants for this struct:

1. `NewFromViper() *HostConfig`
2. `NewFromCue(filepath string) *HostConfig`

Variant 1 creates the struct using `viper.Get()`-like calls to supply the values to the struct. We use this variant when the `--cue` flag is empty.

Otherwise, when a Cue file is specified, variant 2 is used. This means the `schema.cue` file must support default values like the way Viper is set up now. 

I believe this approach simplifies things a lot and the way configs are handled for both formats will be the same code and decoupled.

PR #169, opened 2025-01-16
Cue/cli compatibility fixes

AlanRostem:
Previously, the `--hosts` cli flag was omitted when integrating Cue. Now I have introduced `--client-hosts` and `--replica-hosts` to the cli so that things can work like before and be compatible with Cue if the user chooses to. 

Upon implementing this fix, I see that the choice between Cue and Viper cli configuration can be improved so I opened issue #168

### DIFF
--- a/internal/cli/run.go
+++ b/internal/cli/run.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"bufio"
-	"fmt"
 	"io"
 	"log"
 	"math"
@@ -67,7 +66,8 @@ func init() {
 	runCmd.Flags().StringSlice("modules", nil, "Name additional modules to be loaded.")
 
 	runCmd.Flags().Bool("worker", false, "run a local worker")
-	runCmd.Flags().StringSlice("hosts", nil, "the remote hosts to run the experiment on via ssh")
+	runCmd.Flags().StringSlice("replica-hosts", nil, "the remote hosts to run replicas on via ssh")
+	runCmd.Flags().StringSlice("client-hosts", nil, "the remote hosts to run clients on via ssh")
 	runCmd.Flags().String("exe", "", "path to the executable to deploy and run on remote workers")
 	runCmd.Flags().String("ssh-config", "", "path to ssh_config file to resolve host aliases (defaults to ~/.ssh/config)")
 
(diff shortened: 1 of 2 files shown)


internal/orchestration/controller.go
```diff
@@ -153,7 +153,7 @@ func (e *Experiment) createReplicas(replicaMap config.ReplicaMap) (cfg *orchestr
 			if err != nil {
 				return nil, err
 			}
-			opt.SetTreeOptions(e.hostCfg.BranchFactor, e.hostCfg.TreePositions, e.hostCfg.TreeDelta)
+			// opt.SetTreeOptions(e.hostCfg.BranchFactor, e.hostCfg.TreePositions, e.hostCfg.TreeDelta)
```
meling, line 156: This was deleted in #170. We should just delete it here as well since it will be compatible with #170.

internal/cli/run.go
```diff
@@ -109,49 +110,74 @@ func runController() {
 	numReplicas := viper.GetInt("replicas")
 	numClients := viper.GetInt("clients")
 
+	// Kauri values
```
meling, line 113:
```go
// Tree-related values
```

internal/cli/run.go
```diff
@@ -109,49 +110,74 @@ func runController() {
...
+	branchFactor := viper.GetUint32("bf")
+
 	cfgPath := viper.GetString("cue")
 
+	treeDelta := viper.GetDuration("tree-delta")
```
meling, line 118: These belong together with tree values above (`bf`)

internal/cli/run.go
```diff
@@ -109,49 +110,74 @@ func runController() {
...
+	intTreePos := viper.GetIntSlice("tree-pos")
+
+	replicaHosts := viper.GetStringSlice("replica-hosts")
+	clientHosts := viper.GetStringSlice("client-hosts")
+
 	var cfg *config.HostConfig
 	if cfgPath != "" {
 		cfg, err = config.Load(cfgPath)
```
meling, line 126: I believe `Load` was renamed in the other PR.

internal/cli/run.go
```diff
@@ -109,49 +110,74 @@ func runController() {
...
+
 	cfgPath := viper.GetString("cue")
 
+	treeDelta := viper.GetDuration("tree-delta")
+	intTreePos := viper.GetIntSlice("tree-pos")
+
+	replicaHosts := viper.GetStringSlice("replica-hosts")
+	clientHosts := viper.GetStringSlice("client-hosts")
+
 	var cfg *config.HostConfig
```
meling, line 124: This is now ExperimentConfig

internal/cli/run.go
```diff
@@ -109,49 +110,74 @@ func runController() {
...
+		treePos = tree.DefaultTreePosUint32(numReplicas)
+	} else {
+		treePos = make([]uint32, len(intTreePos))
+		for i, pos := range intTreePos {
+			treePos[i] = uint32(pos)
+		}
+	}
+	if viper.GetBool("random-tree") {
+		rnd := rand.New(rand.NewSource(rand.Uint64()))
+		rnd.Shuffle(len(treePos), reflect.Swapper(treePos))
```
meling, line 172: We now have `tree.Shuffle()` for this.

meling (commented):
I've done a quick pass on this one. It looks like this is incompatible with #170. I suggest we close this one after migrating any important updates to #170. Otherwise, it will be painful to merge either this one or #170.

Issue #150, opened 2025-01-06
Refactor: Simplify the experiment framework


meling:
The current experiment framework in `internal/orchestration` has some methods that are rather complex and it is therefore difficult to adapt to new uses. I'm opening this to point out some concerns with the current code and propose a way to simplify the logic of parts of the `orchestration` package to make the code more maintainable.

In `controller.go`:
- `func (e *Experiment) assignReplicasAndClients() (err error)`: This code is 114 lines long and has a cyclomatic complexity of 16. Simplify by decoupling the logic into two separate methods, one for replicas and one for clients. Part of this code is already implemented in the new `cue-config` branch, which should be merged soon. It may not have the same behavior as before, but it is unclear that the current behavior is what we want.
- `func (e *Experiment) createReplicas() (cfg *orchestrationpb.ReplicaConfiguration, err error)`: This code is 60 lines long. This is not too bad, but it could still be simplified. For instance, the logic to compute the `validFor` slice could be a function.
- The `Experiment` struct stores too much information; it should only store static information provided via the command line or a configuration file. The results from `assignReplicasAndClients` and similar outputs should be kept separate from the input data. In the `config.Config` struct in the `cue-config` branch, I return a `ReplicaMap` defined like this: `type ReplicaMap map[string][]*orchestrationpb.ReplicaOpts`.
- The `HostConfig` struct can be replaced by the `config.Config` struct.
- Both `Experiment` and `HostConfig` contain `NumReplicas = Replicas` and `NumClients = Clients`.
- It is not clear to me if we need the `internal-address`. See below.
- The following fields of `Experiment` should not be in the struct; there are too many maps with the same key.

```go
HostConfigs map[string]HostConfig
Byzantine   map[string]int // number of replicas to assign to each byzantine strategy

// the host associated with each replica.
hostsToReplicas map[string][]hotstuff.ID
// the host associated with each client.
hostsToClients map[string][]hotstuff.ID
replicaOpts    map[hotstuff.ID]*orchestrationpb.ReplicaOpts
```

In `internal/cli/run.go`:
- `func runController()`: This code is 119 lines long.

The `internal-address` question; from `docs/experimentation.md`:

Additionally, it is possible to specify an _internal address_ for each host.
The internal address is used by replicas instead of the address used by the controller.
This is useful if the controller is connecting to the remote hosts using a global address,
whereas the hosts can communicate using local addresses.
The internal address is configured through the configuration file (loaded by the `--config` flag):

```toml
[...]

PR #170, opened 2025-01-17
Simplify config handling and other fixes to orchestration

### LINKED ISSUE #168: Simplify the configuration handling code

AlanRostem:
Fixes #168

Previously, the `--hosts` cli flag was omitted when integrating Cue. Now I have introduced `--client-hosts` and `--replica-hosts` to the cli so that things can work like before and be compatible with Cue if the user chooses to. 

Both Cue and Viper (.toml) are compatible with the framework. 

The code has also been adapted to support the following:

- Multiple config formats
- One standardized `ExperimentConfig` struct that holds all possible parameters
- Simpler code for experiment/test setup, independent of config format
- CLI flags, `.toml`, and `.cue` configs can work in tandem

This PR also solves #150, #159, and #174


internal/cli/run.go
```diff
@@ -149,81 +47,47 @@ func runController() {
...
 	checkf("failed to connect to remote hosts: %v", err)
 
-	if exePath == "" {
-		exePath, err = os.Executable()
+	if cfg.Exe == "" {
+		cfg.Exe, err = os.Executable()
 		checkf("failed to get executable path: %v", err)
 	}
 
+	// TODO: Generate DeployConfig from ExperimentConfig type.
```
meling, line 58: There is a circular dependency if you move the logic to the config package, since `orchestration/controller.go` uses `config.ExperimentConfig` and more from `config`... We should perhaps rethink if we can structure things a bit better between the packages.

internal/orchestration/controller.go
```diff
@@ -153,7 +141,7 @@ func (e *Experiment) createReplicas(replicaMap config.ReplicaMap) (cfg *orchestr
 			if err != nil {
 				return nil, err
 			}
-			opt.SetTreeOptions(e.hostCfg.BranchFactor, e.hostCfg.TreePositions, e.hostCfg.TreeDelta)
+			// opt.SetTreeOptions(e.hostCfg.BranchFactor, e.hostCfg.TreePositions, e.hostCfg.TreeDelta)
```
meling, line 144: Keep or drop?

internal/orchestration/orchestration_test.go
```diff
@@ -11,176 +11,150 @@ import (
...
-		}()
+func run(t *testing.T, cfg *config.ExperimentConfig) {
+	t.Helper()
 
-		err = experiment.Run()
-		if err != nil {
-			t.Fatal(err)
-		}
+	controllerStream, workerStream := net.Pipe()
+	workerProxy := orchestration.NewRemoteWorker(protostream.NewWriter(controllerStream), protostream.NewReader(controllerStream))
```
meling, line 78: I got curious about this line and the next one; we use `controllerStream` in both on the same line, and `workerStream` in both on the next line. Does this make sense?
Looking at code from `cli/run.go`:
```go
		remoteWorkers[host] = orchestration.NewRemoteWorker(
			protostream.NewWriter(session.Stdin()), protostream.NewReader(session.Stdout()),
		)
```
We use a different stream for the reader and writer.

meling (changes requested):

internal/orchestration/controller.go
AlanRostem: This can be droppped since the values are set from `CreateReplicaOpts` in `ExperimentConfig`.

internal/cli/run.go
```diff
@@ -14,132 +13,31 @@ import (
...
-		cfg.TreeDelta = viper.GetDuration("tree-delta")
+	cuePath := viper.GetString("cue")
+	if cuePath != "" {
+		cfg, err = config.NewCue(cuePath, cfg)
+		checkf("config error when loading %s: %v", cuePath, err)
 	}
 
-	worker := viper.GetBool("worker")
+	// A cleaner way would be to have this in the constructor of
+	// config, but this is done here since the Cue might overwrite
```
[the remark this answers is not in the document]
AlanRostem: The above comment refers to both, but then you'd have to duplicate the code for both constructors, which is why I added the comment. An even cleaner approach would be to create a `type TreePositions []uint32` and add a `Randomize()` method to it.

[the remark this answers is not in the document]
AlanRostem: I agree with the "order of magnitude". I will open a separate issue/PR on this

internal/cli/run.go
AlanRostem: Agreed. Perhaps a PR on #172 could contribute to this

meling (approved):

PR #171, opened 2025-01-19
Implemented methods to compute waiting time at the replicas 

hanish520:


kauri/aggduration.go
```diff
@@ -0,0 +1,68 @@
...
+	children := t.tree.ChildrenOf(id)
+	if len(children) == 0 {
+		return 0 // base case: leaf nodes have zero aggregation latency.
+	}
+	// calculate aggregation latencies for each child
+	latencies := make([]time.Duration, len(children))
+	for i, child := range children {
+		latencies[i] = 2*t.lm.Latency(id, child) + t.aggregationLatency(child)
+	}
+	return max(latencies) + t.delta
```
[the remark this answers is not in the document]
hanish520: @leandernikolaus here delta signifies the processing delay at the replica, so I am adding the network latency and processing delay to determine the timer duration at internal nodes in the tree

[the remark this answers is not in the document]
leandernikolaus: @hanish520 nice. Can you add a comment that explains this. 
Is delta on line 60 below also representing the processing delay?

kauri/aggduration.go
```diff
@@ -0,0 +1,68 @@
...
+	"time"
+
+	"github.com/relab/hotstuff"
+	"github.com/relab/hotstuff/internal/latency"
+	"github.com/relab/hotstuff/internal/tree"
+	"github.com/relab/hotstuff/modules"
+)
+
+// AggregationLatency calculates the maximum duration of a path in the tree.
+type AggregationLatency struct {
```
meling, line 14: @hanish520 Wouldn't it be more suitable to keep this in the `tree` package? The aggregation latency is valid for many tree configurations, not just Kauri, no?

kauri/aggduration.go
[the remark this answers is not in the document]
hanish520: sure Leander, the delta at line 60 represents the network + processing delay, this duration computation is used when network latencies are not simulated as we did in Iniva.

kauri/aggduration.go
```diff
@@ -0,0 +1,68 @@
...
+	// calculate aggregation latencies for each child
+	latencies := make([]time.Duration, len(children))
+	for i, child := range children {
+		latencies[i] = 2*t.lm.Latency(id, child) + t.aggregationLatency(child)
+	}
+	return max(latencies) + t.delta
+}
+
+// FixedAggDuration returns the fixed aggregation duration based on the height of the tree.
+func (t *AggregationLatency) fixedAggDuration() time.Duration {
```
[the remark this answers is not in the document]
hanish520: I think so, i will move this to tree module

kauri/aggduration.go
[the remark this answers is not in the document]
leandernikolaus: Ah thanks, I did misunderstand something.

kauri/kauri.go
```diff
@@ -129,8 +131,7 @@ func (k *Kauri) SendProposalToChildren(p hotstuff.ProposeMsg) {
 		}
 		k.logger.Debug("Sending proposal to children ", children)
 		config.Propose(p)
-		// delta is the network delay between two processes, at root it is 2*(3-1)*delta
-		waitTime := time.Duration(2*(k.tree.TreeHeight()-1)) * k.treeConfig.TreeWaitDelta()
+		waitTime := k.tree.WaitTimerDuration(k.server.LatencyMatrix(), k.opts)
 		go k.WaitToAggregate(waitTime, k.currentView)
```
meling, line 135: Here you are passing `k.currentView` as `view` to `WaitToAggregate`, and on Line 114 you compare `k.currentView != view`. Thus, `view` is always going to be equal to `k.currentView`, and this check and return is useless:
```go
if k.currentView != view {
	return
}
```

kauri/kauri.go
```diff
@@ -129,8 +131,7 @@ func (k *Kauri) SendProposalToChildren(p hotstuff.ProposeMsg) {
 		}
 		k.logger.Debug("Sending proposal to children ", children)
 		config.Propose(p)
-		// delta is the network delay between two processes, at root it is 2*(3-1)*delta
-		waitTime := time.Duration(2*(k.tree.TreeHeight()-1)) * k.treeConfig.TreeWaitDelta()
+		waitTime := k.tree.WaitTimerDuration(k.server.LatencyMatrix(), k.opts)
```
meling, line 134: Move the `waitTime` variable into the `WaitToAggregate` method.

internal/tree/treelatency_test.go
```diff
@@ -0,0 +1,72 @@
...
+			timerType = modules.WaitTimerFixed
+		} else {
+			timerType = modules.WaitTimerAgg
+		}
+
+		bf := 2
+		treePos := DefaultTreePos(test.size)
+
+		opts := modules.OptionsWithID(test.id)
+		opts.SetTreeConfig(uint32(bf), treePos, time.Duration(test.delta), timerType)
```
meling, line 64: I don't like the `opts` thing here... It is confusing and too complicated. Couldn't this be done simply by passing the relevant parameters to the `WaitTimerDuration` method?

I also don't like the `timerType` thing. I'm still thinking about a better design.

kauri/kauri.go
hanish520: This check was intended to prevent aggregation from the previous view sent to the parent, k.currentView may change to next view while the parameter remains the same, enabling us to check if the replica is still in the current view.

internal/tree/treelatency_test.go
hanish520: we are passing the delta and timerType values from the experiment config, we can make them part of the replica/server structure and read from there.

kauri/kauri.go
meling: I don't understand this. I don't think this works like you are intending.
1. `k.WaitToAggregate()` is called only from one place.
2. It is called with the parameter `k.currentView` --> `view` (Thus, inside the func, `view` is the same as `k.currentView`)
3. On Line 114, you compare `k.currentView` with `view` (which is always the same as `k.currentView`) -->
4. The if-check is non-functional since it will never trigger an early return.

internal/tree/treelatency.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/relab/hotstuff"
+	"github.com/relab/hotstuff/internal/latency"
+)
+
+// DelayType is the type of delay to use when calculating the time to wait.
+type DelayType int
+
+const (
+	// AggregationTime is the delay type for the time to wait for the aggregation of votes
```
leandernikolaus, line 15:
```suggestion
	// AggregationTime is the delay type for setting a wait time based on the latency of links in the tree.
```

internal/tree/treelatency.go
```diff
@@ -0,0 +1,62 @@
...
+	// based on the latency of each node in the tree.
```
leandernikolaus, line 16:
```suggestion
```

internal/tree/treelatency.go
```diff
@@ -0,0 +1,62 @@
...
+	AggregationTime DelayType = iota
+	// TreeHeightTime is the delay type for the time to wait based on the height of the tree.
```
leandernikolaus, line 18:
```suggestion
	// TreeHeightTime is the delay type for setting a wait time only based on the height of the tree.
```

meling (approved):

Issue #172, opened 2025-01-20
Non-idiomatic Go code


AlanRostem:
In relation to #157 I am opening this issue to start discussing idiomatic Go for this framework.

There are many non-idiomatic Go solutions in the codebase. Particularly, dependency injection is being used to solve for the cyclic dependency issues in the code, which is both non-idiomatic Go and an architectural design flaw. Additionally, the module system heavily uses metaprogramming to assemble and initialize modules. A more idiomatic approach would be to create middleware objects to solve the cyclic dependency issues and explicitly define the dependencies of each module. 

I fiddled with the above concept and it lead to heavily refactoring the codebase. I believe this is a better approach for the long term, but will need some time to get done. 

Please feel free to add some comments. I may be missing other non-idiomatic Go parts of the code

leandernikolaus:
I am not sure what is meant by the middleware approach. 
I think the approach should more be: 
If A and B have a circular dependency, then it might be good to try and extract some aspect from A and B into a separate module, C.

meling:
> If A and B have a circular dependency, then it might be good to try and extract some aspect from A and B into a separate module, C.

Yes, this is the common approach to break circular dependencies. A quick search gave this blog post [Managing Circular Dependencies in Go](https://medium.com/@cosmicray001/managing-circular-dependencies-in-go-best-practices-and-solutions-723532f04dde). I have some books that provide some information about this as well.

Perhaps @AlanRostem you could come up with a proposal for a new package structure, by first analyzing their current dependencies.

Regarding middleware, I'm also uncertain what you mean in this context; people in the gRPC community refer to interceptors as middleware and there is this repo: [go-grpc-middleware](https://github.com/grpc-ecosystem/go-grpc-middleware). The [connect-rpc](https://connectrpc.com) folks provide an interceptor framework for both unary and stream-based RPCs; both incoming requests and outgoing responses and both client- and server-side; not sure if they call it middleware. Folks that create REST-based web frameworks also refer to middleware for certain logical modules that appear sandwiched between other modules... I'm not sure middleware is well-defined, and can thus be a bit confusing.

AlanRostem:
Ah my bad, I meant that middleware is the C in terms of Leander's explanation.

Yes, I'll come up with a proposal upon analyzing the package structure.

PR #179, opened 2025-01-27
Package restructuring to achieve idiomatic Go code

AlanRostem:
Attempts to solve #172

This PR removes the following
- builder for module initialization and dependency injection (and subsequently removes some metaprogramming)
- abstract module definitions
- module registry

...and adds the following
- modules are initialized as soon as they are constructed
- dependency injection is explicit (constructors use explicit struct types) and done in the constructors
- a restructuring of modules/packages to solve cyclic dependencies directly
-  several additional modules/packages that decouple existing functionality to both simplify the code and solve cyclic dependencies
- modules with several possible implementations (like `leaderrotation`) are constructed through a simple `switch`-statement rather than a complex, metaprogrammed registry 

The following is a dependency graph as a result of the above changes. Note that not all packages are included here, but only the ones relevant to cyclic references and consensus logic:

modules drawio(4)

The following is a list of the new modules/packages:

- `certauth`: creates and verifies signatures and certificates. Previously in `crypto`
- `sender`: sends protocol-related messages. Previously in `backend` and was part of `Configuration`
- `committer`: commits valid blocks and checks forks. Previously in `consensus`

Some modules were moved out of existing packages or renamed
- `clientsrv` and `cmdcache` were previously in `backend`
- `backend` renamed to `server`
- TimeoutContext-related tools in `synchronizer` are moved to a separate `synctools` package since many depend on it.

To summarize: cyclic dependencies have been solved directly by decoupling relevant functionality of one or more modules to a separate one, allowing the cyclic dependers to reference one common decoupled dependency. Furthermore, cyclic references have been solved in some areas thanks to parametrization. For instance, some modules would require the synchronizer just to get the view number in one method. Now the method takes a view number instead of having to ref the synchronizer. 

There are still more changes that can be done to decouple some modules where parametrization certainly can help. Also, reducing the complexity of `consensus` and `synchronizer` is desirable.


internal/orchestration/components.go
```diff
@@ -121,24 +121,156 @@ func getCrypto(
...
+	EventLoop *eventloop.EventLoop
+	Logger    logging.Logger
+}
+
+func NewCoreComponents(
+	id hotstuff.ID,
+	logTag string,
+	privKey hotstuff.PrivateKey,
+) *CoreComponents {
+	opts := core.NewOptions(id, privKey)
```
meling, line 135: Stuff like this can be done directly:
```go
return &CoreComponents{
	Options:   core.NewOptions(id, privKey),
	Logger: logging.New(fmt.Sprintf("%s%d", logTag, id)),
	EventLoop: eventloop.New(logger, 100),
}
```

protocol/consensus/consensus.go
```diff
@@ -0,0 +1,250 @@
...
+		configuration: configuration,
+		eventLoop:     eventLoop,
+		logger:        logger,
+		opts:          opts,
+
+		lastVote: 0,
+	}
+
+	cs.eventLoop.RegisterHandler(hotstuff.ProposeMsg{}, func(event any) {
+		cs.OnPropose(cs.view, event.(hotstuff.ProposeMsg))
```
leandernikolaus, line 76: If this is called based on a message, who tells the synchronizer to advance the view?

leandernikolaus (commented):
I think the view can be removed from all signatures but the Propose called by the synchronizer.

The point on how to advance the view when OnPropose is called from the event loop is a major issue. I think you need to instead of relying on a return value add a NewView event to the event loop.

internal/orchestration/components.go
```diff
@@ -0,0 +1,473 @@
...
+				Certificates: []tls.Certificate{certificate},
+				ClientCAs:    rootCAs,
+				ClientAuth:   tls.RequireAndVerifyClientCert,
+			})),
+		))
+	}
+
+	coreComps := NewCoreComponents(opts.HotstuffID(), "hs", privKey)
+	coreComps.Options.SetSharedRandomSeed(opts.GetSharedSeed())
+	// TODO: Upon a merge with master, this doesn't compile.
```
meling, line 409: Should fix reminder.

internal/proto/hotstuffpb/convert.go
```diff
@@ -1,14 +1,14 @@
-// Package hotstuffpb contains protocol buffers message types and conversion functions for the HotStuff protocol.
+// Package convert contains conversion methods between protocol buffer message types and HotStuff protocol message structures.
```
meling, line 1: This is now back in the `hotstuffpb` package. Also, I think they are functions, not methods.

internal/proto/hotstuffpb/convert.go
```diff
@@ -142,7 +142,7 @@ func BlockToProto(block *hotstuff.Block) *Block {
 	}
 }
 
-// BlockFromProto converts a hotstuffpb.Block to a consensus.Block.
+// BlockFromProto converts a Block to a consensus.Block.
```
meling, line 145: This is now `hotstuff.Block`

internal/proto/hotstuffpb/convert.go
```diff
@@ -130,7 +130,7 @@ func ProposalFromProto(p *Proposal) (proposal hotstuff.ProposeMsg) {
 	return
 }
 
-// BlockToProto converts a consensus.Block to a hotstuffpb.Block.
+// BlockToProto converts a consensus.Block to a Block.
```
meling, line 133: `consensus.Block` --> `hotstuff.Block`

protocol/rules/byzantine/byzantine.go
```diff
@@ -3,66 +3,47 @@ package byzantine
...
-	modules.RegisterModule("silence", func() Byzantine { return &silence{} })
-	modules.RegisterModule("fork", func() Byzantine { return &fork{} })
-}
+const (
+	SilenceModuleName = "silence"
+	ForkModuleName    = "fork"
+)
 
 // Byzantine wraps a consensus rules implementation and alters its behavior.
 type Byzantine interface {
```
meling, line 17: Delete interface. It is not necessary anymore.

protocol/rules/byzantine/byzantine.go
```diff
@@ -3,66 +3,47 @@ package byzantine
...
+	modules.ConsensusRules
 }
 
 func (s *silence) ProposeRule(_ hotstuff.SyncInfo, _ hotstuff.Command) (hotstuff.ProposeMsg, bool) {
 	return hotstuff.ProposeMsg{}, false
 }
 
-func (s *silence) Wrap(rules consensus.Rules) consensus.Rules {
-	s.Rules = rules
+func (s *silence) Wrap() modules.ConsensusRules {
```
meling, line 30: Remove `Wrap`.

protocol/rules/byzantine/byzantine.go
```diff
@@ -92,11 +73,20 @@ func (f *fork) ProposeRule(cert hotstuff.SyncInfo, cmd hotstuff.Command) (propos
...
+		ConsensusRules: rules,
+
+		blockChain: blockChain,
+		opts:       opts,
+	}
 }
 
-func (f *fork) Wrap(rules consensus.Rules) consensus.Rules {
-	f.Rules = rules
+func (f *fork) Wrap() modules.ConsensusRules {
```
meling, line 90: Remove `Wrap`.

protocol/rules/byzantine/byzantine.go
```diff
@@ -92,11 +73,20 @@ func (f *fork) ProposeRule(cert hotstuff.SyncInfo, cmd hotstuff.Command) (propos
...
 
 // NewFork returns a byzantine replica that will try to fork the chain.
-func NewFork(rules consensus.Rules) consensus.Rules {
-	return &fork{Rules: rules}
+func NewFork(
+	rules modules.ConsensusRules,
+
+	blockChain *blockchain.BlockChain,
+	opts *core.Options,
+) Byzantine {
```
meling, line 81: Should return `*fork` since it already embeds the provided `rules`.

protocol/rules/byzantine/byzantine.go
```diff
@@ -3,66 +3,47 @@ package byzantine
...
-func (s *silence) Wrap(rules consensus.Rules) consensus.Rules {
-	s.Rules = rules
+func (s *silence) Wrap() modules.ConsensusRules {
 	return s
 }
 
 // NewSilence returns a byzantine replica that will never propose.
-func NewSilence(c consensus.Rules) consensus.Rules {
-	return &silence{Rules: c}
+func NewSilence(rules modules.ConsensusRules) Byzantine {
```
meling, line 35: Should return `*silence` since it already embeds the provided `rules`.

meling: We should have a test for these Byzantine rules to check that they do what we expect them to do...

meling (changes requested):
Lots of great stuff here... I've covered less than half of it.

protocol/rules/byzantine/byzantine.go
```diff
@@ -3,66 +3,47 @@ package byzantine
...
-		mod.InitModule(mods)
-	}
+	blockChain *blockchain.BlockChain
+	opts       *core.Options
+	modules.ConsensusRules
 }
 
-func (f *fork) ProposeRule(cert hotstuff.SyncInfo, cmd hotstuff.Command) (proposal hotstuff.ProposeMsg, ok bool) {
-	block, ok := f.blockChain.Get(f.synchronizer.HighQC().BlockHash())
+func (f *fork) ProposeRule(view hotstuff.View, highQC hotstuff.QuorumCert, cert hotstuff.SyncInfo, cmd hotstuff.Command) (proposal hotstuff.ProposeMsg, ok bool) {
```
leandernikolaus, line 45: I think you can use `syncinfo.QC()` instead of passing the `highQC`. This allows you to get rid of the `highQC` also in the signature of `Propose`.

protocol/synchronizer/synchronizer.go
```diff
@@ -35,28 +45,49 @@ type Synchronizer struct {
...
+		auth:          auth,
+		configuration: configuration,
+		sender:        sender,
+		eventLoop:     eventLoop,
+		logger:        logger,
+		opts:          opts,
+
+		currentView: 1,
+
+		timer:    oneShotTimer{time.AfterFunc(0, func() {})}, // dummy timer that will be replaced after start() is called
```
leandernikolaus, line 85: Do you need this? Previously the timer was not explicitly initialized.

protocol/synchronizer/synchronizer.go
```diff
@@ -338,12 +362,15 @@ func (s *Synchronizer) AdvanceView(syncInfo hotstuff.SyncInfo) {
...
 
 	s.logger.Debugf("advanced to view %d", newView)
-	s.eventLoop.AddEvent(ViewChangeEvent{View: newView, Timeout: timeout})
+	s.eventLoop.AddEvent(hotstuff.ViewChangeEvent{View: newView, Timeout: timeout})
 
 	leader := s.leaderRotation.GetLeader(newView)
 	if leader == s.opts.ID() {
-		s.consensus.Propose(syncInfo)
-	} else if replica, ok := s.configuration.Replica(leader); ok {
+		sInfo, ok := s.consensus.Propose(s.View(), s.highQC, syncInfo)
```
leandernikolaus, line 369: I think the return value here can actually be omitted. 
Passing back the same QC will not do anything new. 
The important part is to receive the QC when OnPropose is called from the network. See comments in consensus.go.

AlanRostem:
# Further Restructuring

I have compiled additional future work needed for this PR to be merged. 

## Component Bundling

All cases like this in `internal/orchestration/components.go` below should be moved to respective packages, since right now the code is really big for one file. For example, the below code can be moved to `core/components.go` since each component is in the same root directory. Same can be done with the other packages like `protocol`, where the consensus and synchronizer bundling can be moved to a `protocol/components.go` where the package is named `protocol` and its dependencies are `consensus`, `synchronizer`, `byzantine`, etc.


```go
type CoreComponents struct {
	Options   *core.Options
	EventLoop *eventloop.EventLoop
	Logger    logging.Logger
}

func NewCoreComponents(
	id hotstuff.ID,
	logTag string,
	privKey hotstuff.PrivateKey,
) *CoreComponents {
	logger := logging.New(fmt.Sprintf("%s%d", logTag, id))
	return &CoreComponents{
		Options:   core.NewOptions(id, privKey),
		EventLoop: eventloop.New(logger, 100),
		Logger:    logger,
	}
}
```

(**Suggestion**: Rename `<name>Components` structs to `<name>Bundle`.) 

This is necessary since many tests, like twins and go tests, will depend on bundles of components in the future. Thus, using the `internal/orchestration` package is inconsistent. 

## Network Component Changes

The objects that interact *directly* (i.e., send/receive messages) with the network layer are in the following files:

- `network/sender/replica.go`
- `network/sender/sender.go`
- `service/clientsrv/clientsrv.go`
- `service/server/server.go`

In general, we should move the logical aspects of these files to separate objects, preserve their networking code, and inject the logical objects as dependencies to these networked componentse. For example, `clientsrv.go` should have its execution and fork handling logic decoupled, and reference a separate component which handles that logic.

We should move `server.go` to the `network` directory and rename it `receiver.go` so it matches the same pattern as done with `sender.go`. It also relates to the network layer, so it should be bundled with `NetworkComponents`. 

For `replica.go`, this should be moved to `internal/orchestration` since its logic is only used there now. I believe it was an interface since it was being used by twins before, but creating a replica (i.e., initializing needed components for each use cases) should be unique to each scenario.

### A Network Layer Module 

After the above changes, it would be very convenient for creating various tests and scenarios if the networking part (sending and receiving messages) was modular. Just like how `Consensus` needs a `modules.ConsensusRules` interface (with a unique implementation) to function, a similar thing could be done for the `sender.Sender`. That way, we can create self-contained tests with an emulated network quite easily. 

I am not sure how I would do it with a reciever (`server.Server`), but I imagine we can utilize the exisitng `hotstuffpb.HotStuff` in `internal/proto/hotstuffpb/hotstuff_gorums.pb.go` and embed it into a `Receiver` interface. If we do this, we must ensure `kauripb.Kauri` from `internal/proto/kauripb/kauri_gorums.pb.go`, and similar services is a base class for consistency's sake. But regardless, we run into the issue where our gorums config is coupled with component logic. Looking at the code now, there is no dependency on the receiver, and events can simply be passed to the core eventloop manually if one creates a unique testing scenario with an emulated network. In general, receiving messages is done by popping out of the eventloop, so a component for receiving is not strictly necessary for testing, but having the general logic for one would be nice when considering induced latency and tests revolving around that.

## Further Decoupling

Shouldn't `sender.Sender` be decoupled from `blockchain.BlockChain`? The only place inside the blockchain the sender is being used is `BlockChain.Get`, where if a block is not found locally it is fetched from another replica. I think this logic is needed in many cases, but not all. Decoupling this and fetching the block manually outside of `Get` is more consistent and less complex. Thus, the blockchain's purpose is simply a local storage of blocks. But, I am not sure how easy this is to do since I am not familiar with the reason for its coupling. I think we should think about this when restructuring several parts of `consensus` and `synchronizer`.

## Re-Creating Tests

This PR has broken the majority of the tests since a lot of them used mocking, which is something that got omitted. With the above suggestions and usage of component bundling, I think re-creating tests should be simple. I recommend that we look into further decoupling, such as with the taking the sender out of blockchain, to simplify the bundles even further. An good example of this re-creation is in `security/certauth/certauth_test.go`, where the crypto stuff is easily testable using concrete and simple bundles of components. I still think this test can be simplified further by decoupling the sender from blockchain, but this is a good start for such a concept.

internal/orchestration/components.go
```diff
@@ -0,0 +1,468 @@
...
+	Multiplier   float64
+}
+
+func NewProtocolModules(
+	coreComps *CoreComponents,
+	netComps *NetworkComponents,
+	secureComps *SecurityComponents,
+	srvComps *ServiceComponents,
+
+	opts *orchestrationpb.ReplicaOpts, // TODO: avoid modifying this so it doesn't depend on orchestrationpb
```
meling, line 260: I'm not sure I understand the TODO, could you clarify.

Also, remove blank line above.

internal/orchestration/components.go
```diff
@@ -0,0 +1,468 @@
...
+	consensusName, leaderRotationName, byzantineName string,
+	vdOpt ViewDurationOptions,
+) (*ProtocolModules, error) {
+	consensusRules, err := getConsensusRules(consensusName, secureComps.BlockChain, coreComps.Logger, coreComps.Options)
+	if err != nil {
+		return nil, fmt.Errorf("invalid consensus name: '%s'", consensusName)
```
meling, line 266: Should return `err` directly, since the same error message is returned from `getConsensusRules`.

internal/orchestration/components.go
```diff
@@ -0,0 +1,468 @@
...
+		// create tree only if we are using tree leader (Kauri)
+		coreComps.Options.SetTree(createTree(opts))
+		duration = synchronizer.NewFixedViewDuration(opts.GetInitialTimeout().AsDuration())
+	} else {
+		duration = synchronizer.NewViewDuration(
+			vdOpt.SampleSize, vdOpt.StartTimeout, vdOpt.MaxTimeout, vdOpt.Multiplier,
+		)
+	}
+	if byzantineName != "" {
+		if byz, err := getByzantine(byzantineName, consensusRules, secureComps.BlockChain, coreComps.Options); err == nil {
```
meling, line 293: Here we should do:
```go
if byz, err := getByzantine(byzantineName ...); err != nil {
	return nil, err
}
consensusRules = byz
coreComps.Logger.Infof("assigned byzantine strategy: %s", byzantineName)
```
Rename also: `byzantineName` to `byzantineStrategy`

internal/orchestration/components.go
```diff
@@ -0,0 +1,468 @@
...
+
+	coreComps := NewCoreComponents(opts.HotstuffID(), "hs", privKey)
+	coreComps.Options.SetSharedRandomSeed(opts.GetSharedSeed())
+	// TODO: Upon a merge with master, this doesn't compile.
+	// coreComps.Options.SetTreeConfig(opts.GetBranchFactor(), opts.TreePositionIDs(), opts.TreeDeltaDuration())
+
+	netComps := NewNetworkComponents(
+		coreComps,
+		creds,
+		gorums.WithDialTimeout(opts.GetConnectTimeout().AsDuration()),
```
meling, line 410: This is deprecated and no longer supported in gorums.

internal/orchestration/components.go
AlanRostem: If we want to move this struct and its corresponding constructor to a `security/components.go`, then it should not depend on the `orchestrationpb.ReplicaOpts` since it's complex in comparison. This makes it hard to create tests etc. 

I think this `opts` exists because of the following line and the one below it: https://github.com/relab/hotstuff/blob/4736b1413293479ffe5eafcb3f388b93ef5d3b3f/internal/orchestration/components.go#L283

...but after fixing the todo below it which you wrote, this could be removed easily.

meling (changes requested):
Posting those I have in my stack...

metrics/registry.go
```diff
@@ -12,18 +19,24 @@ var (
...
+	eventLoop *eventloop.EventLoop,
+	logger logging.Logger,
+	metricsLogger Logger,
+	opts *core.Options,
+	measurementInterval time.Duration,
+	names ...string) {
+	for _, name := range names {
+		switch name {
+		case NameClientLatency:
+			NewClientLatency(eventLoop, logger, opts, metricsLogger)
```
meling, line 32: Shouldn't these constructors be recorded somewhere?

internal/proto/hotstuffpb/convert_test.go
```diff
@@ -78,6 +78,8 @@ func TestConvertBlock(t *testing.T) {
 }
 
 // TODO(meling): disabled since it fails after the package-restructuring and I haven't been able to fix it yet.
+// TODO(AlanRostem): discovered the bug - the metadata per replica is nil and is needed by BLS12 to generate public keys.
```
meling, line 81: Great! On that note, [Gorums v0.9.0](https://github.com/relab/gorums/releases/tag/v0.9.0) now supports passing [metadata per message](https://github.com/relab/gorums/blob/ad120bcf1ed7628fe12f9b083894f38f8992259f/tests/metadata/metadata_test.go#L101) as well, not only per connection.

metrics/registry.go
AlanRostem: It's not needed since the constructors add handlers to the eventloop, so the GC will keep them in memory. Plus, none of the objects have public methods so everything is handled internally by each object. 

I would suggest not returning the objects and rename the constructors to `InitClientLatency` for instance since the return value is not used.

AlanRostem: This approach should be better

```go
func Enable(
	eventLoop *eventloop.EventLoop,
	logger logging.Logger,
	metricsLogger Logger,
	opts *core.Options,
	measurementInterval time.Duration,
	names ...string) {
	for _, name := range names {
		switch name {
		case NameClientLatency:
			EnableClientLatency(eventLoop, logger, opts, metricsLogger)
		case NameViewTimeouts:
			EnableViewTimeouts(eventLoop, logger, metricsLogger, opts)
		case NameThroughput:
			EnableThroughput(eventLoop, logger, metricsLogger, opts)
		}
	}
	AddTicker(eventLoop, measurementInterval)
}
```

internal/orchestration/worker.go
```diff
@@ -238,7 +264,7 @@ func (w *Worker) startClients(req *orchestrationpb.StartClientRequest) (*orchest
 			RateStepInterval: opts.GetRateStepInterval().AsDuration(),
 			Timeout:          opts.GetTimeout().AsDuration(),
 		}
-		buildOpt := core.NewOptions(hotstuff.ID(opts.GetID()), nil)
+		buildOpt := core.NewGlobals(hotstuff.ID(opts.GetID()), nil, 0) // TODO(AlanRostem): can seed be zero?
```
meling: Why not set the seed via a option method `WithSharedRandomSeed`?

meling: What should the default be, if it is not set?

core/globals.go
```diff
@@ -0,0 +1,120 @@
...
+	shouldVerifyVotesSync bool
+
+	sharedRandomSeed   int64
+	connectionMetadata map[string]string
+
+	tree          tree.Tree
+	shouldUseTree bool
+}
+
+func NewGlobals(id hotstuff.ID, pk hotstuff.PrivateKey, seed int64, opts ...GlobalsOption) *Globals {
```
meling: Should this panic if `id` is 0 and similar if `pk` is nil?

meling (changes requested):
This is a review of a specific commit; I've looked at several individual commits, which didn't have any issues.

service/cmdcache/options.go
```diff
@@ -0,0 +1,12 @@
+package cmdcache
+
+type Option func(*Cache)
+
+func WithBatching(batchSize uint32) Option {
+	if batchSize == 0 {
+		panic("batch size must be higher than zero")
```
meling, line 7: greater than.

Do we want to panic if batch size is 1? We should probably not use `WithBatching(1)`, but there could be an argument made for allowing it to avoid having to remove the entire option when using the code in a loop with `WithBatching(k)`...

service/cmdcache/options.go
```diff
@@ -0,0 +1,12 @@
...
+	}
+	return func(cmdCache *Cache) {
+		cmdCache.batchSize = int(batchSize)
```
meling, line 10: Could we make `cmdCache.batchSize` an `uint32`?

internal/orchestration/worker.go
AlanRostem: I think the default is already zero in the master branch, but I'm unsure if that's intentional since it's Go's default value initialization of ints. 

Why do we want shared seed in the first place? I can imagine it's useful for debug purposes (maybe), otherwise by default I would use the UNIX now time for instance.

core/globals.go
AlanRostem: Yes, I think so. ìd=0` is invalid and so is `pk=nil`.

AlanRostem: Nevermind, `pk` is being set to nil when initializing a client: https://github.com/relab/hotstuff/blob/ccf209f59a3823d8ac8378f611f89c394ef55416/internal/orchestration/worker.go#L288

Even on master branch: https://github.com/relab/hotstuff/blob/461e32c91b5664b825e445c8749e835944d1413f/internal/orchestration/worker.go#L329

Not sure if this makes sense...

protocol/consensus/consensus.go
```diff
@@ -0,0 +1,114 @@
...
+	err := cs.voter.Verify(proposal)
+	if err != nil {
+		cs.logger.Errorf("failed to verify incoming vote: %v", err)
+		return
+	}
+	// if we can't commit the block yet, don't vote for it.
+	if !cs.committer.TryCommit(block) {
+		return
+	}
+	// even if voting fails, we should be able to go to the next view.
```
leandernikolaus, line 75: I believe the vote below should only fail, if there is some error with the crypto setup.
So for protocol flow I believe voting failing is irrelevant.

protocol/consensus/consensus.go
```diff
@@ -0,0 +1,114 @@
...
+func (cs *Consensus) OnPropose(proposal *hotstuff.ProposeMsg) {
+	block := proposal.Block
+	// ensure that I can vote in this view based on the protocol's rule.
+	err := cs.voter.Verify(proposal)
+	if err != nil {
+		cs.logger.Errorf("failed to verify incoming vote: %v", err)
+		return
+	}
+	// if we can't commit the block yet, don't vote for it.
+	if !cs.committer.TryCommit(block) {
```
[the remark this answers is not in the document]
leandernikolaus: Does the Accept change any state? If not it should be in the Verify above.

protocol/consensus/consensus.go
```diff
@@ -0,0 +1,114 @@
...
+		cs.OnPropose(&p)
+	})
+	return cs
+}
+
+// OnPropose is called when receiving a proposal from a leader and returns true if the proposal was voted for.
+func (cs *Consensus) OnPropose(proposal *hotstuff.ProposeMsg) {
+	block := proposal.Block
+	// ensure that I can vote in this view based on the protocol's rule.
+	err := cs.voter.Verify(proposal)
```
leandernikolaus, line 66: This verify contains different parts.
1. It verifies that the proposal is well formed (crypto and from correct leader)
2. It verifies that we can vote for it (vote rule and last voted)
It is a change from previous logic, if blocks that pass 1 but not 2 are not handled for committing and advancing highQC. If these blocks are not stored in the blockchain, it may be necessary to later fetch them. 
However, this should happen only in problematic edge cases so not sure performance (additional fetch) is an issue here.

protocol/consensus/consensus.go
```diff
@@ -0,0 +1,114 @@
...
+		SyncInfo: newInfo,
+	})
+	// try to vote for the block and retrieve its partial certificate.
+	pc, err := cs.voter.Vote(block)
+	// don't send the vote if it failed
+	if err != nil {
+		cs.logger.Errorf("failed to vote: %v", err)
+		return
+	}
+	cs.extHandler.OnPropose(proposal, pc)
```
leandernikolaus, line 88: I do not understand what this is doing. I also looked at the interface and none of the comments there help me understand. Probably this needs different names.

protocol/consensus/consensus.go
```diff
@@ -0,0 +1,114 @@
...
+	proposal, err := cs.proposer.CreateProposal(view, highQC, syncInfo)
+	if err != nil {
+		// NOTE: errors frequently come from this method, so we just debug log here.
+		cs.logger.Debugf("could not create proposal: %v", err)
+		return
+	}
+	block := proposal.Block
+	// as proposer, I can vote for my own proposal without verifying.
+	// NOTE: this vote call is not likely to fail since the leader does it, otherwise something went terribly wrong...
+	pc, err := cs.voter.Vote(block)
```
leandernikolaus, line 102: I believe this should result in some issues, since you are jumping over the side effects in tryCommit.

protocol/proposer/proposer.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// CreateProposal attempts to create a new outgoing proposal if a command exists and the protocol's rule is satisfied.
+func (cs *Proposer) CreateProposal(view hotstuff.View, highQC hotstuff.QuorumCert, syncInfo hotstuff.SyncInfo) (proposal hotstuff.ProposeMsg, err error) {
+	err = nil
+	ctx, cancel := timeout.Context(cs.eventLoop.Context(), cs.eventLoop)
+	defer cancel()
+	// find a value to propose.
+	// NOTE: this is blocking until a batch is present in the cache.
+	cmd, ok := cs.commandCache.Get(ctx)
```
leandernikolaus, line 47: If you get commands, do you still need to mark them as proposed, e.g. call `proposed(cmd)` or is that included in `Get`?

protocol/consensus/consensus.go
leandernikolaus: If you run this code in single leader mode, does the leader ever commit anything?

leandernikolaus (commented):
Some comments on the new consensus.

protocol/consensus/consensus.go
leandernikolaus: @meling can you read this?

protocol/consensus/consensus.go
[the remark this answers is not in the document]
AlanRostem: Hmm, no it doesn't, so this can be moved into `Verify`. Though, bloats `voter` with more dependencies, but maybe the `committer` shouldnt have `cmdcache` anyway?

protocol/consensus/consensus.go
AlanRostem: Agreed. It should just continue, and maybe log the error

protocol/consensus/consensus.go
AlanRostem: Just tested it and it does not! Is it enough just to add a `Commit` before the `Vote` here to fix the issue?

protocol/proposer/proposer.go
AlanRostem: Yes, it's done later in `TryCommit`

https://github.com/relab/hotstuff/blob/98f52d0f9dda186d3693dc55bdcf38e33a0d3bce/service/committer/committer.go#L75

but I think this line can be moved into `Consensus` after  `CreateProposal` succeeded

service/cmdcache/cmdcache.go
```diff
@@ -1,4 +1,4 @@
-package replica
+package cmdcache
```
meling, line 1: I don't think we need a separate package for this. Logically, I wonder if it would make sense to move this `Cache` to the `clientpb` or `clientsrv` package?

Right now, I feel we have gone too far and added too many packages.

service/cmdcache/cmdcache.go
```diff
@@ -58,13 +73,13 @@ func (c *cmdCache) addCommand(cmd *clientpb.Command) {
 }
 
 // Get returns a batch of commands to propose.
```
meling, line 75: This comment doesn't appear correct, but it's the same on master.

meling: I see that the logic actually takes a batch of commands and marshals them into a string (hotstuff.Command)... It's a bit strange IMO.

service/cmdcache/cmdcache.go
meling: Another name for this concept that's become common these days is a mempool. Maybe that's what we should call the `clientpb` package?

service/clientsrv/clientsrv.go
```diff
@@ -9,51 +9,51 @@ import (
...
-func (srv *clientSrv) InitModule(mods *modules.Core) {
-	mods.Get(
-		&srv.eventLoop,
-		&srv.logger,
-	)
-	srv.cmdCache.InitModule(mods)
-}
-
-func (srv *clientSrv) Start(addr string) error {
+func (srv *Server) Start(addr string) error {
```
meling, line 56: This is never called.

service/clientsrv/clientsrv.go
```diff
@@ -71,25 +71,40 @@ func (srv *clientSrv) StartOnListener(lis net.Listener) {
...
 }
 
-func (srv *clientSrv) Stop() {
+func (srv *Server) Stop() {
 	srv.srv.Stop()
 }
 
-func (srv *clientSrv) ExecCommand(ctx gorums.ServerCtx, cmd *clientpb.Command) (*emptypb.Empty, error) {
-	id := cmdID{cmd.ClientID, cmd.SequenceNumber}
+func (srv *Server) CmdCache() *cmdcache.Cache {
```
meling, line 78: Never called

service/clientsrv/clientsrv.go
```diff
@@ -1,4 +1,4 @@
-package replica
+package clientsrv
```
meling, line 1: This suggestion might be controversial. Maybe even this `clientsrv` package can be moved into the current `clientpb` package (assuming that we rename it to `mempool`). Then this file would be named `mempool_server.go` and the struct would be accessed as `mempool.Server` and `mempool.New` or `mempool.NewServer`. (see also my comments in `cmdcache.go`)

meling (changes requested):
I've added some comments... still not finished (this will take a while, I suspect.)

dependencies/consensus.go
```diff
@@ -0,0 +1,59 @@
+package dependencies
```
meling, line 1: This package name is a bit awkward and long imo. ChatGPT suggests a few names: `builder`, `wiring`, `compose`, `bootstrap` and more... I tend to like `wiring` best among these because `builder` is not quite representative for what this does. I feel builder is something more low-level, while `wiring` gives an indication that these are pre-wired dependencies. Another alternative is `bundle`, but not sure `bundle.NewConsensus` is better than `wiring.NewConsensus`... Maybe `wire.NewConsensus` is better??

service/clientsrv/clientsrv.go
```diff
@@ -114,17 +132,19 @@ func (srv *clientSrv) Exec(cmd hotstuff.Command) {
...
+func (srv *Server) Fork(cmd hotstuff.Command) {
 	batch := new(clientpb.Batch)
 	err := proto.UnmarshalOptions{AllowPartial: true}.Unmarshal([]byte(cmd), batch)
 	if err != nil {
 		srv.logger.Errorf("Failed to unmarshal command: %v", err)
 		return
 	}
-
 	for _, cmd := range batch.GetCommands() {
 		srv.mut.Lock()
```
meling, line 143: We can move the lock after creating the `id`.

core/config.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// PrivateKey returns the private key.
+func (g *RuntimeConfig) PrivateKey() hotstuff.PrivateKey {
+	return g.privateKey
+}
+
+// HasAggregateQC returns true if aggregated quorum certificates should be used.
+// This is true for Fast-Hotstuff: https://arxiv.org/abs/2010.11454
+func (g *RuntimeConfig) HasAggregateQC() bool {
```
meling, line 52: Can we remove this option and just let it be set by fast-hotstuff.

meling:
Closing this since I've upgraded this to be the new `main` (default) branch.

PR #182, opened 2025-01-28
refactor(consensus): split onPropose

leandernikolaus:
Added `check` and `createVote` functions
which do not have side effects and should be easier to test.

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -166,14 +166,59 @@ func (cs *consensusBase) Propose(cert hotstuff.SyncInfo) {
 }
 
 func (cs *consensusBase) OnPropose(proposal hotstuff.ProposeMsg) { //nolint:gocyclo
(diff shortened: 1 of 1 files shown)


consensus/consensus.go
```diff
@@ -166,14 +166,59 @@ func (cs *consensusBase) Propose(cert hotstuff.SyncInfo) {
...
+	if !cs.check(proposal) {
+		return
+	}
+
+	block := proposal.Block
+
+	if qcBlock, ok := cs.blockChain.Get(block.QuorumCert().BlockHash()); ok {
+		cs.acceptor.Proposed(qcBlock.Command())
+	} else {
+		cs.logger.Info("OnPropose: Failed to fetch qcBlock")
```
leandernikolaus, line 180: I believe this case can never happen since the existence of the qcBlock was already checked when verifying the QC.
Should we remove the whole if?

PR #187, opened 2025-02-17
ConsensusLatency: Add consensus latency metrics

hanish520:


types.go
```diff
@@ -383,3 +384,7 @@ func writeParticipants(wr io.Writer, participants IDSet) (err error) {
 	})
 	return err
 }
+
+type ConsensusLatencyMeasurementEvent struct {
```
meling, line 388: Suggest to rename to `ConsensusLatencyEvent`

metrics/consensuslatency.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/relab/hotstuff/modules"
+)
+
+func init() {
+	RegisterReplicaMetric("consensus-latency", func() any {
+		return &ConsensusLatency{}
+	})
+}
+
+// ConsensusLatency processes ConsensusLatencyMeasurementEvent, and
```
meling, line 19:
```go
// ConsensusLatency processes consensus latency measurements and writes them to the metrics logger.
```

metrics/consensuslatency.go
```diff
@@ -0,0 +1,70 @@
...
+// InitModule gives the module access to the other modules.
+func (lr *ConsensusLatency) InitModule(mods *modules.Core) {
+	var (
+		eventLoop *eventloop.EventLoop
+		logger    logging.Logger
+	)
+
+	mods.Get(
+		&lr.metricsLogger,
+		&lr.opts,
```
meling, line 37: There is only one use of `lr.opts` on Line 63: `uint32(lr.opts.ID())`. We could instead make `id` be part of the struct and initialize it in `InitModule` from a local field `opts`. This would save us from having `opts` be part of the struct.

block.go
```diff
@@ -72,6 +75,11 @@ func (b *Block) View() View {
 	return b.view
 }
 
+// TS returns the timestamp of the block
+func (b *Block) TS() time.Time {
```
meling, line 79: Suggest to rename it to `Timestamp` and change the comment to:
```go
// Timestamp returns the block's timestamp.
```

meling (changes requested):
Minor changes suggested @hanish520

metrics/consensuslatency.go
```diff
@@ -0,0 +1,70 @@
...
+// writes ConsensusLatencyMeasurementEvent to the metrics logger.
+type ConsensusLatency struct {
+	metricsLogger Logger
+	opts          *modules.Options
+
+	wf Welford
+}
+
+// InitModule gives the module access to the other modules.
+func (lr *ConsensusLatency) InitModule(mods *modules.Core) {
```
AlanRostem, line 29: Please create a similar function: https://github.com/relab/hotstuff/blob/44b43fe960255e19ec37af9963df6cc17b66a1bb/metrics/throughput.go#L26

...and pass the event loop and logger into the function. You may do the same with the ID to avoid keeping opts.

consensus/consensus.go
```diff
@@ -282,6 +283,7 @@ func (cs *consensusBase) commitInner(block *hotstuff.Block) error {
 	} else {
 		return fmt.Errorf("failed to locate block: %s", block.Parent())
 	}
+	cs.eventLoop.AddEvent(hotstuff.ConsensusLatencyMeasurementEvent{Latency: time.Since(block.TS())})
```
AlanRostem, line 286: `commitInner` was moved to the `committer`, so this line should be here: https://github.com/relab/hotstuff/blob/44b43fe960255e19ec37af9963df6cc17b66a1bb/service/committer/committer.go#L100

AlanRostem (changes requested):
Please see my review comments... Merging `package-restructuring` into this branch is necessary to apply the changes I requested.

meling:
@AlanRostem My idea would be to merge this into `master` and then merge `master` into `package-restructuring`. Do you see any issues with that?

AlanRostem:
> @AlanRostem My idea was to merge this into `master` and then merge `master` into `package-restructuring`. Do you see any issues with that?

Ah, I misunderstood. Yes this is doable as long as we address the changes I requested upon merging master into package-restructuring. 

Though, I am still wondering about the block timestamps I mentioned in my other comment...

block.go
```diff
@@ -25,6 +27,7 @@ func NewBlock(parent Hash, cert QuorumCert, cmd Command, view View, proposer ID)
 		cmd:      cmd,
 		view:     view,
 		proposer: proposer,
+		ts:       time.Now(),
```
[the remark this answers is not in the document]
meling: The `Block` is sent on the wire like this:
```proto
message Block {
  bytes Parent = 1;
  QuorumCert QC = 2;
  uint64 View = 3;
  bytes Command = 4;
  uint32 Proposer = 5;
}
```
So this `ts` field is not sent over the wire; I believe it is a local-only timestamp.

@hanish520 Please confirm my understanding. I also suggest to rename the field to `timestamp`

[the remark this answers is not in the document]
AlanRostem: I see. So the timestamp is recorded here: 
https://github.com/relab/hotstuff/blob/04497a2842ad354be61829060f719da4a3b28e9f/block.go#L30

but `NewBlock` is used to convert from proto to the `Block`:

https://github.com/relab/hotstuff/blob/461e32c91b5664b825e445c8749e835944d1413f/internal/proto/hotstuffpb/convert.go#L149

As I understand, it measures "local commit latency". I.e., it measures the latency between block arrival and commit if the replica is an acceptor. For leader, this latency measurement is very tiny, no? @hanish520 is this the intention?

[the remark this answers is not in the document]
hanish520: This metrics is used to measure the commit latency at the leader when used in the network simulation mode. This metrics may not make sense in the normal mode.

meling (approved):

PR #192, opened 2025-05-25
Added timestamp to the block

hanish520:


internal/proto/hotstuffpb/hotstuff.proto
```diff
@@ -6,6 +6,8 @@ import "gorums.proto";
 
 import "google/protobuf/empty.proto";
```
meling, line 8: Nit: Let's keep import lines tight.

block.go
```diff
@@ -34,6 +34,12 @@ func NewBlock(parent Hash, cert QuorumCert, cmd Command, view View, proposer ID)
 	return b
 }
 
+func (b *Block) SetTimestamp(ts time.Time) {
```
meling, line 37: While touching this file, let's rename `TimeStamp` on Line 85 to the correct spelling: `Timestamp`

genesis.go
```diff
@@ -1,8 +1,12 @@
 package hotstuff
 
+import "time"
+
 var genesisBlock = NewBlock(Hash{}, QuorumCert{}, "", 0, 0)
 
 // GetGenesis returns a pointer to the genesis block, the starting point for the hotstuff blockchain.
 func GetGenesis() *Block {
+	ts := time.Unix(0, 0).UTC() // should we use a static string for parsing?
```
meling, line 9: I think we should do the timestamp initialization either via `NewBlock` or in an `init()` function so that we don't incur any construction cost when calling `GetGenesis`, even though we rarely call it.

We could probably use a timestamp newer than the Unix zero value (1970) time... Something like 2025 Jan 1, but it is not important, I think.

PR #198, opened 2025-05-30
Mock sender for testing, remove logger from crypto modules, making more tests, decouple client from consensus, fix twins tests

AlanRostem:
The mock sender implements the `modules.Sender` interface and allows for creating simple unit tests for classes that need a sender (i.e., `consensus.Proposer`). All invocations to the mock-sender's methods (i.e., `Propose`) will store the message in a slice which can be retrieved later in the test. The idea is to check if a message was sent after some consensus logic to pass the test.


meling:
Had a quick glance! I like it... (replacing `ok bool` with `error` returns and avoiding the logger ...)

security/crypto/bls12/bls12.go
```diff
@@ -217,46 +216,40 @@ func (bls *bls12Base) coreVerify(pubKey *PublicKey, message []byte, signature *b
 	return engine.Result().IsOne()
 }
 
-func (bls *bls12Base) popProve() *bls12.PointG2 {
+func (bls *bls12Base) popProve() (*bls12.PointG2, error) {
 	pubKey := bls.privateKey().Public().(*PublicKey)
 	proof, err := bls.coreSign(pubKey.ToBytes(), domainPOP)
 	if err != nil {
-		bls.logger.Panicf("Failed to generate proof-of-possession: %v", err)
+		return nil, fmt.Errorf("Failed to generate proof-of-possession: %v", err)
```
meling, line 223: lower case: failed ... and use the `%w` verb instead of `%v` when adding `err` at the end like this, because then the root cause error can be unwrapped from the returned error. (see the std lib errors package for detailed docs.)

PS: This rule only applies to error messages, not logging.

security/crypto/bls12/bls12.go
```diff
@@ -217,46 +216,40 @@ func (bls *bls12Base) coreVerify(pubKey *PublicKey, message []byte, signature *b
...
-			bls.logger.Warnf("Invalid proof-of-possession for replica %d", replica.ID)
-		}
-	}()
-
+func (bls *bls12Base) checkPop(replica *hotstuff.ReplicaInfo) (err error) {
 	popBytes, ok := replica.Metadata[popMetadataKey]
 	if !ok {
-		bls.logger.Warnf("Missing proof-of-possession for replica: %d", replica.ID)
-		return false
+		return fmt.Errorf("Missing proof-of-possession for replica: %d", replica.ID)
```
meling, line 235: lower case: missing...

security/crypto/bls12/bls12.go
```diff
@@ -217,46 +216,40 @@ func (bls *bls12Base) coreVerify(pubKey *PublicKey, message []byte, signature *b
...
 }
 
-func (bls *bls12Base) checkPop(replica *hotstuff.ReplicaInfo) (valid bool) {
-	defer func() {
-		if !valid {
-			bls.logger.Warnf("Invalid proof-of-possession for replica %d", replica.ID)
-		}
-	}()
-
+func (bls *bls12Base) checkPop(replica *hotstuff.ReplicaInfo) (err error) {
```
meling, line 232: We should avoid naming the return types unless we use them in the code or they serve to document the method signature. We should also be conservative about using named return types in code, because there are some common pitfalls with doing so... (search the internet or ChatGPT for details...)

security/crypto/bls12/bls12.go
```diff
@@ -350,75 +343,85 @@ func (bls *bls12Base) Combine(signatures ...hotstuff.QuorumSignature) (combined
 			}
 			g2.Add(&agg, &agg, &sig2.sig)
 		} else {
```
meling, line 345: If you flip the `ok` check on Line 332 to `!ok`, you could move this error return early and avoid the else part so that the main logic can be unindented for a more readable flow.

security/crypto/bls12/bls12.go
```diff
@@ -350,75 +343,85 @@ func (bls *bls12Base) Combine(signatures ...hotstuff.QuorumSignature) (combined
...
-			bls.logger.Panicf("cannot combine incompatible signature type %T (expected %T)", sig1, sig2)
+			return nil, fmt.Errorf("cannot combine incompatible signature type %T (expected %T)", sig1, sig2)
```
meling, line 346: Maybe these types of errors can be made a `FatalError` since we actually want to log with a `Panic` for these things... I think these panics here mean that there is something wrong with the configuration, and the user should fix it.

But we want to return an error here rather than log things at this level, since we want to have tests for this stuff, and testing for errors is easier than testing for errors in logs.

security/crypto/bls12/bls12.go
```diff
@@ -350,75 +343,85 @@ func (bls *bls12Base) Combine(signatures ...hotstuff.QuorumSignature) (combined
...
 		id := firstParticipant(s.Participants())
-		pk, ok := bls.publicKey(id)
-		if !ok {
-			bls.logger.Warnf("Missing public key for ID %d", id)
-			return false
+		pk, err := bls.publicKey(id)
+		if err != nil {
+			return err
+		}
+		if !bls.coreVerify(pk, message, &s.sig, domain) {
```
meling, line 367: If you make `coreVerify` return the error, you can still do it directly like before:
```go
return bls.coreVerify(pk, message, &s.sig, domain)
```

security/crypto/bls12/bls12.go
```diff
@@ -350,75 +343,85 @@ func (bls *bls12Base) Combine(signatures ...hotstuff.QuorumSignature) (combined
...
-		return false
+		pks = append(pks, pk)
+		return true
 	})
-	if len(pks) != n {
-		return false
+	if errs != nil {
+		return fmt.Errorf("one or more public keys had errors: %v", errs)
+	}
+	if !bls.fastAggregateVerify(pks, message, &s.sig) {
```
meling, line 387: Suggest to move the error inside the verify method and return directly:
```go
return bls.fastAggregateVerify(pks, message, &s.sig)
```

security/crypto/bls12/bls12.go
```diff
@@ -350,75 +343,85 @@ func (bls *bls12Base) Combine(signatures ...hotstuff.QuorumSignature) (combined
...
 		}
-		bls.logger.Warnf("Missing public key for ID %d", id)
-		return false
+		pks = append(pks, pk)
+		return true
 	})
-	if len(pks) != n {
-		return false
+	if errs != nil {
+		return fmt.Errorf("one or more public keys had errors: %v", errs)
```
meling, line 385: I think this should be something like: `missing one or more public keys: %w`

security/crypto/bls12/bls12.go
```diff
@@ -350,75 +343,85 @@ func (bls *bls12Base) Combine(signatures ...hotstuff.QuorumSignature) (combined
...
+		pk, err := bls.publicKey(id)
+		if err != nil {
+			return err
 		}
 		pks = append(pks, pk)
 	}
 
 	if len(batch) == 1 {
-		return bls.coreVerify(pks[0], msgs[0], &s.sig, domain)
+		if !bls.coreVerify(pks[0], msgs[0], &s.sig, domain) {
```
meling, line 417: Direct return

security/crypto/bls12/bls12.go
```diff
@@ -350,75 +343,85 @@ func (bls *bls12Base) Combine(signatures ...hotstuff.QuorumSignature) (combined
...
+func (bls *bls12Base) BatchVerify(signature hotstuff.QuorumSignature, batch map[hotstuff.ID][]byte) error {
 	s, ok := signature.(*AggregateSignature)
 	if !ok {
-		bls.logger.Panicf("cannot verify incompatible signature type %T (expected %T)", signature, s)
+		return fmt.Errorf("cannot verify incompatible signature type %T (expected %T)", signature, s)
 	}
 
 	if s.Participants().Len() != len(batch) {
-		return false
+		return fmt.Errorf("expected %d participants", len(batch))
```
meling, line 401: Not sure about this one, but maybe something like this:
`signature mismatch: %s participants, expected: %s`

security/crypto/bls12/bls12.go
```diff
@@ -350,75 +343,85 @@ func (bls *bls12Base) Combine(signatures ...hotstuff.QuorumSignature) (combined
...
 	if len(batch) == 1 {
-		return bls.coreVerify(pks[0], msgs[0], &s.sig, domain)
+		if !bls.coreVerify(pks[0], msgs[0], &s.sig, domain) {
+			return fmt.Errorf("core-verify failed")
+		}
+		return nil
 	}
 
-	return bls.aggregateVerify(pks, msgs, &s.sig)
+	if !bls.aggregateVerify(pks, msgs, &s.sig) {
```
meling, line 423: direct return

security/crypto/ecdsa/ecdsa.go
```diff
@@ -115,80 +110,86 @@ func (ec *ecdsaBase) Combine(signatures ...hotstuff.QuorumSignature) (hotstuff.Q
...
+func (ec *ecdsaBase) Verify(signature hotstuff.QuorumSignature, message []byte) (err error) {
 	s, ok := signature.(crypto.Multi[*Signature])
 	if !ok {
-		ec.logger.Panicf("cannot verify signature of incompatible type %T (expected %T)", signature, s)
+		return fmt.Errorf("cannot verify signature of incompatible type %T (expected %T)", signature, s)
 	}
 	n := signature.Participants().Len()
 	if n == 0 {
-		return false
+		return fmt.Errorf("expected more than zero participants")
```
meling, line 127: `signature mismatch: no participants`

meling: Or maybe:
`verify failed: no participants`

security/crypto/ecdsa/ecdsa.go
```diff
@@ -115,80 +110,86 @@ func (ec *ecdsaBase) Combine(signatures ...hotstuff.QuorumSignature) (hotstuff.Q
...
+func (ec *ecdsaBase) BatchVerify(signature hotstuff.QuorumSignature, batch map[hotstuff.ID][]byte) (err error) {
 	s, ok := signature.(crypto.Multi[*Signature])
 	if !ok {
-		ec.logger.Panicf("cannot verify signature of incompatible type %T (expected %T)", signature, s)
+		return fmt.Errorf("cannot verify signature of incompatible type %T (expected %T)", signature, s)
 	}
 	n := signature.Participants().Len()
 	if n == 0 {
-		return false
+		return fmt.Errorf("expected more than zero participants")
```
meling, line 154: batchverify failed: no participants

security/crypto/ecdsa/ecdsa.go
```diff
@@ -115,80 +110,86 @@ func (ec *ecdsaBase) Combine(signatures ...hotstuff.QuorumSignature) (hotstuff.Q
...
+func (ec *ecdsaBase) verifySingle(sig *Signature, hash hotstuff.Hash) error {
 	replica, ok := ec.config.ReplicaInfo(sig.Signer())
 	if !ok {
-		ec.logger.Warnf("ecdsaBase: got signature from replica whose ID (%d) was not in the config.", sig.Signer())
-		return false
+		return fmt.Errorf("ecdsaBase: got signature from replica whose ID (%d) was not in the config.", sig.Signer())
 	}
 	pk := replica.PubKey.(*ecdsa.PublicKey)
-	return ecdsa.Verify(pk, hash[:], sig.R(), sig.S())
+	if !ecdsa.Verify(pk, hash[:], sig.R(), sig.S()) {
```
meling, line 189: direct return

security/crypto/ecdsa/ecdsa.go
```diff
@@ -115,80 +110,86 @@ func (ec *ecdsaBase) Combine(signatures ...hotstuff.QuorumSignature) (hotstuff.Q
...
+	return nil
 }
 
-func (ec *ecdsaBase) verifySingle(sig *Signature, hash hotstuff.Hash) bool {
+func (ec *ecdsaBase) verifySingle(sig *Signature, hash hotstuff.Hash) error {
 	replica, ok := ec.config.ReplicaInfo(sig.Signer())
 	if !ok {
-		ec.logger.Warnf("ecdsaBase: got signature from replica whose ID (%d) was not in the config.", sig.Signer())
-		return false
+		return fmt.Errorf("ecdsaBase: got signature from replica whose ID (%d) was not in the config.", sig.Signer())
```
meling, line 186: Is it enough to just use `ecdsa:` instead of `ecdsaBase`?
What about:
```go
return fmt.Errorf("ecdsa: received signature from replica ID %d not in config", sig.Signer())
```
(note also, we don't use punctuation in error messages)

security/crypto/eddsa/eddsa.go
```diff
@@ -94,79 +90,83 @@ func (ed *eddsaBase) Combine(signatures ...hotstuff.QuorumSignature) (hotstuff.Q
...
+func (ed *eddsaBase) BatchVerify(signature hotstuff.QuorumSignature, batch map[hotstuff.ID][]byte) (err error) {
 	s, ok := signature.(crypto.Multi[*Signature])
 	if !ok {
-		ed.logger.Panicf("cannot verify signature of incompatible type %T (expected %T)", signature, s)
+		return fmt.Errorf("cannot verify signature of incompatible type %T (expected %T)", signature, s)
 	}
 	n := signature.Participants().Len()
 	if n == 0 {
-		return false
+		return fmt.Errorf("expected more than zero participants")
```
meling, line 133: see above

meling (changes requested):
I did a deeper review now... mostly Go technical details about error messages, simplifications and style etc.

Note I didn't comment on all instances of the same problems, so great if you also check the rest of the changes herein for the same issues.

There are several opportunities for direct returns by moving the error return into more verify methods that currently return bool. However, I looked at after completing the review, and it was not clear which is the best solution... it could be relevant to keep things as a bool as is now, but I did notice this one in `coreVerify`:
```go
	messagePoint, err := g2.HashToCurve(message, domainTag)
	if err != nil {
		return false
	}
```
Which loses its `err` information...

PR #203, opened 2025-06-20
Embed ProposeRuler in consensus.Ruleset

meling:
This embeds the ProposerRuler interface in the consensus.Ruleset (previous modules.HotstuffRuleset). This allows to implement the ProposeRule in the various protocol implementations, avoiding custom logic for AggregateQC handling for protocols that don't need this.

- **feat(consensus): embed ProposeRuler into Ruleset and their impl**
- **refactor(consensus): remove ProposerOption and simplify NewProposer**
- **refactor: move modules.HotstuffRuleset to consensus.Ruleset**
- **refactor: simplify ProposeRule implementations**


AlanRostem (approved):
Love this change! Reviewed everything and it's good to merge.

Just some sidenotes: I noticed in the `modules` package that the consensus rules were moved to `consensus`. We should do a similar thing for other modules too (in a different PR). Also, `registry.go` and `registry_test.go` are not needed anymore and can be deleted.

PR #221, opened 2025-10-03
feat: replace LastVote with expanded StopVoting method

meling:
This adds a bool value to the StopVoting method so that we can avoid accessing the LastVote value only for checking if the prev and current view are different outside the voter.

### DIFF
--- a/protocol/consensus/voter.go
+++ b/protocol/consensus/voter.go
@@ -21,7 +21,7 @@ type Voter struct {
 	auth      *cert.Authority
 	committer *Committer
 
-	lastVote hotstuff.View
+	lastVotedView hotstuff.View
 }
 
 func NewVoter(
@@ -42,7 +42,7 @@ func NewVoter(
 		auth:      auth,
 		committer: committer,
 
-		lastVote: 0,
+		lastVotedView: 0,
 	}
 	return v
 }
@@ -71,8 +71,8 @@ func (v *Voter) OnValidPropose(proposal *hotstuff.ProposeMsg) (errs error) {
 // Returns true if voting was stopped for this view, and false if voting
 // had already been stopped for this view.
 func (v *Voter) StopVoting(view hotstuff.View) bool {
-	if v.lastVote < view {
-		v.lastVote = view
+	if v.lastVotedView < view {
+		v.lastVotedView = view
 		return true
 	}
 	return false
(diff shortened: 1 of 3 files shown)


protocol/consensus/voter.go
```diff
@@ -67,11 +67,15 @@ func (v *Voter) OnValidPropose(proposal *hotstuff.ProposeMsg) (errs error) {
 	return errs
 }
 
-// StopVoting ensures that no voting happens in a view earlier than `view`.
-func (v *Voter) StopVoting(view hotstuff.View) {
-	if v.lastVote < view {
-		v.lastVote = view
+// StopVoting marks this and earlier views as unavailable for voting.
```
meling, line 70: Maybe we should keep this text like:
```go
// StopVoting marks views earlier than the given view as unavailable for voting.
OR
// StopVoting marks views earlier than view as unavailable for voting.
```

protocol/synchronizer/synchronizer.go
meling, line 172: Should this be `nextView`?

protocol/synchronizer/synchronizer.go
```diff
@@ -183,12 +183,9 @@ func (s *Synchronizer) OnLocalTimeout() {
...
 	s.lastTimeout = timeoutMsg
-	// stop voting for current view
-	prev := s.voter.LastVote()
-	s.voter.StopVoting(view)
-	// check if view is the same to log vote stop
-	if prev != view {
-		s.logger.Debugf("Stopped voting in view %d and changed to view %d", prev, view)
+
+	if s.voter.StopVoting(view) {
+		s.logger.Debugf("Stopped voting for view %d", view)
```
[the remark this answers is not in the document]
leandernikolaus: No, at this point, the replica is actually in between views.
It stopped voting in one view but did not start the next view. 
It will only start the next view once it has sufficient timeouts.
We should create an issue to examine this further. I remember a related discussion on PBFT, where PBFT stopped voting once a view timed out but later work only stopped voting once a viewChange message is sent.
Does the synchronizer algorithm implemented here specify that we have to stop voting on a local timeout?

protocol/synchronizer/synchronizer.go
leandernikolaus: I believe here, `view` is the current view which is timed out.

leandernikolaus (commented):
Mostly good, there is an issue about stopping to vote

protocol/synchronizer/synchronizer.go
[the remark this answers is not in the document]
meling: I don't know the answer to that question, but that's what it does.

Issue #249, opened 2025-10-31
bug: Safety is violated


leandernikolaus:
In the presence of an attacker or a twin, replicas can commit different blocks on different branches.

This is shown by the `TestSafetyWithTwins` test on `TwinsIssue` branch.
The issue is probably due to a wrong implementation of the commit rule for chaines hotstuff. 
The commit rule was implemented verbatim from the hotstuff paper. 
However, the paper assumes that dummy blocks are inserted for every view, such that the view of the parent always is the current view-1.
However, our implementation does not use dummy blocks. Instead, normally the parent's view is the same as the view in the QC.

PR #250, opened 2025-10-31
Fixing issues with twins and create scenario for safety violation

leandernikolaus:
This pull request introduces several significant improvements to the twins simulation.
Main changes are:
* Ensure twins use the same private key. This way, their messages are interchangeable for other replicas.
* Update the `NodeID` scheme used in twins. `NetworkID` is no longer unique. Instead, `NetworkID` is 0 for non-twins and 1 or 2 for twins.
* Update the network to follow a global view. Messages sent in views higher than the global view are delayed. Once no more messages are delivered, the global view is updated.

This PR should be merged after #246.

### DIFF
--- a/twins/network.go
+++ b/twins/network.go
@@ -21,7 +21,7 @@ import (
 // Replicas without twin have TwinID 0.
 type NodeID struct {
 	ReplicaID hotstuff.ID
-	TwinID uint32
+	TwinID    uint32
 }
 
 func (id NodeID) String() string {
@@ -38,7 +38,7 @@ func (id NodeID) Twin(twinID uint32) NodeID {
 func Replica(id hotstuff.ID) NodeID {
 	return NodeID{
 		ReplicaID: id,
-		TwinID: uint32(0),
+		TwinID:    uint32(0),
 	}
 }
 
@@ -95,9 +95,9 @@ func NewSimpleNetwork(numNodes int) *Network {
 	return network
 }
 
-// NewPartitionedNetwork creates a new Network with the specified list of views.  
-// Each view must specify a leader and a set of partitions, each with a set of nodes.  
-// One or more message types may be specified, which will be dropped at the sending node.  
+// NewPartitionedNetwork creates a new Network with the specified list of views.
+// Each view must specify a leader and a set of partitions, each with a set of nodes.
+// One or more message types may be specified, which will be dropped at the sending node.
 func NewPartitionedNetwork(views []View, dropTypes ...any) *Network {
 	n := &Network{
 		nodes:      make(map[NodeID]*node),
@@ -117,7 +117,8 @@ func NewPartitionedNetwork(views []View, dropTypes ...any) *Network {
 // Twins receive the same private key.
 func (n *Network) createNodesAndTwins(nodes []NodeID, consensusName string) error {
 	for _, nodeID := range nodes {
-		var privKey *ecdsa.PrivateKey; var err error
+		var privKey *ecdsa.PrivateKey
+		var err error
 		if twins := n.replicas[nodeID.ReplicaID]; len(twins) == 0 {
 			// generate new key since this is the first replica with this ReplicaID
 			privKey, err = keygen.GenerateECDSAPrivateKey()
@@ -135,8 +136,7 @@ func (n *Network) createNodesAndTwins(nodes []NodeID, consensusName string) erro
 		n.nodes[nodeID] = node
 		n.replicas[nodeID.ReplicaID] = append(n.replicas[nodeID.ReplicaID], node)
 	}
-	
-	
+
 	// need to configure the replica info after all of them were set up
 	for _, node := range n.nodes {
 		config := node.config
--- a/twins/scenario_test.go
+++ b/twins/scenario_test.go
@@ -36,7 +36,7 @@ func TestPartitionedScenario(t *testing.T) {
 	t.Logf("Network log:\n%s", result.NetworkLog)
 }
 
-// TestPartitionedScenario2 checks that chained HotStuff will commit a block 
+// TestPartitionedScenario2 checks that chained HotStuff will commit a block
 // in fourth view after a partition
 // all nodes are honest and the leader is in a separate partition.
 func TestPartitionedScenario2(t *testing.T) {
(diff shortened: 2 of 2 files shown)


leandernikolaus:
The update in commit 28e9915 should fix #249. Safety test is passing now. 
Need to do some more testing.

protocol/rules/chainedhotstuff.go
```diff
@@ -70,7 +70,11 @@ func (hs *ChainedHotStuff) CommitRule(block *hotstuff.Block) *hotstuff.Block {
 		return nil
 	}
 
-	if block1.Parent() == block2.Hash() && block2.Parent() == block3.Hash() {
+	// since our implementation does not create dummy blocks every view, 
+	// we explicitly check that the parents are in the previous view
```
meling: Do you mean both parent and grandparent here? The check below only checks the view of block1 against block3. But perhaps that is enough since the block hashes matches. If so, we should add a comment about that.

twins/network.go
```diff
@@ -77,10 +95,11 @@ func NewSimpleNetwork(numNodes int) *Network {
 // partitions specifies the network partitions for each view.
 func NewPartitionedNetwork(views []View, dropTypes ...any) *Network {
```
meling: May not be related to any changes herein, but I wonder if we could improve the doc comment here. What about?:
```go
// NewPartitionedNetwork creates a new Network with the specified list of views.
// Each view must specify a leader and a set of partitions, each with a set of nodes.
// One or more message types may be specified, which will be dropped at the sending node.
```

twins/network.go
```diff
@@ -11,41 +11,58 @@ import (
 
 	"github.com/relab/hotstuff"
 	"github.com/relab/hotstuff/core/logging"
+	"github.com/relab/hotstuff/security/crypto/keygen"
 )
 
 // NodeID is an ID that is unique to a node in the network.
 // The ReplicaID is the ID that the node uses when taking part in the consensus protocol,
 // while the NetworkID is used to distinguish nodes on the network.
 type NodeID struct {
```
meling: We should add a doc comment to explain that `TwinID` values are expected. E.g., is `TwinID = 0` ==> no twin, normal replica, while `TwinID = 1` is the first twin, and `TwinID = 2` is the second twin of three replicas?

twins/network.go
```diff
@@ -11,41 +11,58 @@ import (
...
-	NetworkID uint32
+	TwinID uint32
 }
 
 func (id NodeID) String() string {
-	return fmt.Sprintf("r%dn%d", id.ReplicaID, id.NetworkID)
+	return fmt.Sprintf("r%dn%d", id.ReplicaID, id.TwinID)
+}
+
+func (id NodeID) Twin(twinId uint32) NodeID {
```
meling: Could be nice to have a doc comment to explain expected/legal values.

twins/generator_test.go
```diff
@@ -47,3 +47,33 @@ func TestPartitionSizes(t *testing.T) {
...
+
+func TestAssignNodeIDs(t *testing.T) {
+	tests := []struct {
+		name      string
+		numNodes  uint8
+		numTwins  uint8
+		wantNodes []NodeID
+		wantTwins []NodeID
+	}{
+		{"no twins", 3, 0, []NodeID{{1, 0}, {2, 0}, {3, 0}}, nil},
```
meling: Should we update these to use the new API for creating replicas and twins? I'm okay with the current approach, but if keep it we should add a comment about the ordering and meaning of the replica ID vs twin ID.

twins/node.go
```diff
@@ -164,3 +160,10 @@ func newNode(n *Network, nodeID NodeID, consensusName string) (*node, error) {
 	}
 	return node, nil
 }
+
+func (n *node) EffectiveView() hotstuff.View {
```
meling: Not sure I quite understand the meaning of effective here. Could we add a doc comment?

The node has an `effectiveView` that is different from the node's `viewStates.View()`?

twins/generator.go
```diff
@@ -24,29 +24,19 @@ type Generator struct {
...
-				ReplicaID: replicaID,
-				NetworkID: networkID,
-			})
-			networkID++
-			twins = append(twins, NodeID{
-				ReplicaID: replicaID,
-				NetworkID: networkID,
-			})
+			// twins get networkID 1 and 2
+			twins = append(twins, RepID(replicaID).Twin(1))
```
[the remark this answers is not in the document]
leandernikolaus: Fixed in [85f0c45](https://github.com/relab/hotstuff/pull/250/commits/85f0c450aee92b89a99bbefa1bb4a0781cb1af12)

twins/generator_test.go
leandernikolaus: I feel adding `Replica(1)` instead of `{1,0}` makes it harder to read.
Also, is this test not enough documentation in itself? 
I feel it nicely shows the use of these.
If any, there should be documentation in the `assignNodeIDs` function.

twins/network.go
```diff
@@ -61,13 +78,14 @@ type Network struct {
 func NewSimpleNetwork(numNodes int) *Network {
 	allNodesSet := make(NodeSet)
 	for i := 1; i <= numNodes; i++ {
-		allNodesSet.Add(uint32(i))
+		allNodesSet.Add(NodeID{ReplicaID: hotstuff.ID(i), TwinID: uint32(i)})
```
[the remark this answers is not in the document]
leandernikolaus: Yes, good catch. It seems this is currently not used. The tests that do use it ignore the twinID.

twins/network.go
leandernikolaus: Very nice. Thanks

twins/network.go
```diff
@@ -90,12 +109,25 @@ func NewPartitionedNetwork(views []View, dropTypes ...any) *Network {
...
 
 func (n *Network) createTwinsNodes(nodes []NodeID, consensusName string) error {
+	lastReplicaID := hotstuff.ID(1)
+	lastPrivKey, err := keygen.GenerateECDSAPrivateKey()
+	if err != nil {
+		return fmt.Errorf("failed to generate private key: %w", err)
+	}
 	for _, nodeID := range nodes {
-		node, err := newNode(n, nodeID, consensusName)
+		if nodeID.ReplicaID != lastReplicaID {
```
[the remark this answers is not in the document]
leandernikolaus: Hi, not sure I agree this is simpler. 
However, do you have an idea how to do this without relying on that twins are consecutive nodes with the same replicaID?

PR #279, opened 2025-11-16
fixed delayuntil in kauri's begin

### LINKED ISSUE #267: bug: kauri's DelayUntil provide a func() without a corresponding event handler registration
Kauri's Begin method has a [DelayUntil](https://github.com/relab/hotstuff/blob/48e4d893397a306ab3fa2183ac2474f1a83fdb05/kauri/kauri.go#L97) call (on the master branch) with a `func()` as the event type. However, no event handler is registered for such an event type elsewhere in the code.

The code now looks like this (to be pushed to a new PR soon):
```go
	if !k.initDone {
		eventloop.DelayUntil[network.ConnectedEvent](k.eventLoop, func() {
			if err := k.begin(p, pc); err != nil {
				k.logger.Error(err)
			}
		})
		return nil
	}
```
However, there is nothing that will trigger this `func()` event, as it won't match any previously registered event handlers.

The fix should be to add a call to Register with the appropriate handler logic. We currently have these handlers registered:
```go
	eventloop.Register(el, func(_ hotstuff.ReplicaConnectedEvent) {
		k.initDone = true // signal that we are connected
	})
	eventloop.Register(el, func(event kauri.ContributionRecvEvent) {
		k.onContributionRecv(event)
	})
	eventloop.Register(el, func(event WaitTimerExpiredEvent) {
		k.onWaitTimerExpired(event)
	})
```
Currently, it is unclear to me how to resolve this issue, but network/sender.go contains similar functionality for handling delayed connection processing.

More details from ChatGPT Codex:

Because the payload in this case is a zero-arg function literal, the intent is to re-enqueue a callback that restarts begin(p, pc) once the network is ready. The closure captures p and pc, so, in theory, when it is finally executed, it would retry the start-up path using exactly the same arguments as the failed attempt.

However, the event loop only executes handlers that were explicitly registered for a given type. For the closure to run, we would need a handler registered for the func() type (e.g., eventloop.Register(el, func(cb func()) { cb() })). A quick inspection of the repository doesn’t show any such registration, so enqueuing this bare function currently just results in an event that nobody handles—in other words, the retry code never fires. If we really want this pattern, we need to add a handler for func() events (or wrap the retry logic in a proper event struct that already has a handler).

hanish520:
Fixes #267


meling (changes requested):
I wonder if it would be possible to get Copilot to prepare some test cases for the Kauri implementation. Perhaps guiding it to use the `clique_test.go` as a basis? Could you give it a go @hanish520?

meling (approved):
Looks good to me!

PR #287, opened 2025-11-30
chore: remove outdated TODO comment in VerifyQuorumCert

SherlockShemol:
The TODO was marked as 'FIX BUG' but the bug has already been fixed. The code correctly checks for nil signature and returns an error. Changed to a descriptive comment explaining why the check exists.


meling:
The TODO comment is there because we want to find a better solution to this problem so that we don't get nil here. We previously did not have to check for nil.

PR #294, opened 2026-01-12
initial commit for new client

hanish520:
This is a draft PR for new client design. This PR is created to reduce the load on the server to send response to all commands in a client. The design is not final, servers need to send QC along with the command status to justify it. Working on those changes.


client/client.go
```diff
@@ -257,6 +280,46 @@ loop:
 	return nil
 }
 
+func (c *Client) fetchCommandStatus(sequenceNumber uint64) hotstuff.CommandStatus {
+	ticker := time.NewTicker(100 * time.Millisecond)
```
meling: This code repeats the `CommandStatus` call several times, losing the benefit of separating `ExecCommand` (multicast dissemination) from the `CommandStatus` call (result checking). Removing the loop and the ticker is the right approach here. Something like this worked for me:
```go
func (c *Client) fetchCommandStatus(sequenceNumber uint64) hotstuff.CommandStatus {
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	// ctx, cancel := context.WithTimeout(context.Background(), c.timeout)
	defer cancel()

	cmd := &clientpb.Command{
		ClientID:       uint32(c.id),
		SequenceNumber: sequenceNumber,
	}
	response, err := c.gorumsConfig.CommandStatus(ctx, cmd)
	if err != nil {
		c.logger.Errorf("Failed to fetch command status (client: %d, sequence: %d): %v", c.id, sequenceNumber, err)
		return hotstuff.UNKNOWN
	}
	if response == nil || response.Command == nil {
		c.logger.Errorf("Invalid response received when fetching command status (client: %d, sequence: %d)", c.id, sequenceNumber)
		return hotstuff.UNKNOWN
	}
	c.logger.Infof("Fetched command status (client: %d, sequence: %d, status: %d)", c.id, sequenceNumber, response.Status)
	status := hotstuff.CommandStatus(response.Status)
	if status == hotstuff.COMMITTED || status == hotstuff.EXECUTED || status == hotstuff.FAILED {
		return status
	}
	return hotstuff.UNKNOWN
}
```
That is, we use a timeout context to cancel the call if the quorum call didn't finish in time. Note that I tried using `c.timeout` instead of 5 seconds, but the default `c.timeout` is only 500 ms, which was too short for this call.

Remember that the quorum call is also processed in the `CommandStatusQF`, and we should normally receive a quorum of replies without having to repeat the call multiple times. If we don't get a reply, it's likely the replicas are overloaded, and repeating the call here won't help.

client/client.go
```diff
@@ -274,24 +337,24 @@ func (c *Client) handleCommands(ctx context.Context) (executed, failed, timeout
...
-		if err != nil {
-			if errors.Is(err, context.DeadlineExceeded) {
-				c.logger.Debug("Command timed out.")
-				timeout++
-			} else if !errors.Is(err, context.Canceled) {
-				c.logger.Debugf("Did not get enough replies for command: %v\n", err)
-				failed++
-			}
-		} else {
+		response := c.fetchCommandStatus(cmd.sequenceNumber)
```
meling: Another note about performance. We are now actually calling `ExecCommand` followed by a `CommandStatus` quorum call, essentially doubling the "effort" that the replicas have to deal with.

I thought the idea was to use `ExecCommand` for many client commands in a loop, and then periodically (e.g., every 5 seconds) call `CommandStatus` to ask the replicas to provide "proof of execution" for a sequence of the client's commands. (We can simplify the "proof of execution" to just send back the range for the `ClientID` that the replicas have executed... and later we can add proper proofs.)

client/client.go
```diff
@@ -31,17 +31,37 @@ type qspec struct {
 	faulty int
 }
 
-func (q *qspec) ExecCommandQF(_ *clientpb.Command, signatures map[uint32]*emptypb.Empty) (*emptypb.Empty, bool) {
-	if len(signatures) < q.faulty+1 {
+func (q *qspec) CommandStatusQF(command *clientpb.Command, replies map[uint32]*clientpb.CommandStatusResponse) (*clientpb.CommandStatusResponse, bool) {
```
meling, line 64: See also my comment below; I think this should be redesigned to keep track of how many commands each replica has executed and combine the results into a single reply that tells the client how many of its commands have actually been executed. For example, something along these lines (I'm not checking for nil or other failure modes here):
```go
lowMin, highMin := uint64(0), uint64(math.MaxUint64)
for _, resp := range replies {
	if resp.LowSequenceNumber != command.LowSequenceNumber {
		// invalid reply??
	}
	lowMin = min(lowMin, resp.LowSequenceNumber)
	highMin = min(highMin, resp.HighSequenceNumber)
}
```
The logic above isn't quite right, since we need f+1 "proofs" of the `highMin` value.

internal/proto/clientpb/client.proto
```diff
@@ -12,11 +12,27 @@ service Client {
...
+        option (gorums.multicast) = true;
   }
+
+  rpc CommandStatus(Command) returns (CommandStatusResponse) {
+         option (gorums.quorumcall) = true; 
+  }
+}
+
+message CommandStatusResponse {
+  enum Status {
```
meling: I don't think we need the status enum. We are not really interested in unknown, pending, or failed. I'm also not sure we need to distinguish between committed and executed. I think we only need one concept (executed), and then we can remove the enum. (See my other suggestions regarding the high and low sequence number thing above.)

Regarding "failed", this should only happen if a replica crashed, in which case we should get an error from the `CommandStatus` call made to that replica, or no response. This would be filtered in the quorum function.

client/client.go
meling: You probably want to handle errors something like this:
```go
if errors.Is(err, context.DeadlineExceeded) {
	c.logger.Debug("Command timed out.")
	timeout++
} else if !errors.Is(err, context.Canceled) {
	c.logger.Debugf("Did not get enough replies for command: %v\n", err)
	failed++
}
```

meling (changes requested):
I haven't looked at the `clientio.go` code much yet. Leaving it for the next round.

